PR #79, opened 2014-05-16
Proof of concept X-Poedit-Potfile header

joostdekeijzer:
Hello,

When using Poedit, I usually have projects with pot files.

With the "Update from POT File" I always have to navigate to the pot file (which is always in the same location for a project).

I decided I'd try a proof-of-concept X-Poedit-Potfile header storing the location of the pot file. See changes (I do not regularly program C...)

I've only tested this on MacOS X with a relative path (and no BaseDir set).

What's your opinion?

### DIFF
--- a/src/catalog.cpp
+++ b/src/catalog.cpp
@@ -396,6 +396,7 @@ void Catalog::HeaderData::ParseDict()
     // Parse extended information:
     SourceCodeCharset = GetHeader("X-Poedit-SourceCharset");
     BasePath = GetHeader("X-Poedit-Basepath");
+    PotFile = GetHeader("X-Poedit-Potfile");
 
     Keywords.Clear();
     wxString kwlist = GetHeader("X-Poedit-KeywordsList");
@@ -955,6 +956,8 @@ Catalog::Catalog()
 {
     m_isOk = true;
     m_header.BasePath = wxEmptyString;
+    m_header.PotFile = wxEmptyString;
+
     for(int i = BOOKMARK_0; i < BOOKMARK_LAST; i++)
     {
         m_header.Bookmarks[i] = -1;
@@ -1675,6 +1678,16 @@ bool Catalog::Update(ProgressInfo *progress, bool summary, bool *cancelledByUser
             wxLogError(_("Source code directory '%s' doesn't exist."), path.c_str());
             return false;
         }
+
+        wxString pot_file = m_header.PotFile;
+        if (!pot_file.empty())
+        {
+            if (!wxIsAbsolutePath(pot_file))
+                pot_file = path + "/" + pot_file;
+    
+            return UpdateFromPOT(pot_file, summary, cancelledByUser);
+        }
+
 
         wxSetWorkingDirectory(path);
     }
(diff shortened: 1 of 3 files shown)


mtancoigne:
Hi!

**TL;DR:**
  - I agree with @joostdekeijzer that selecting the source file on every update is a pain (i'm now _cleaning_ my translations, so i update the .pot file on every small change in my sources), so it would be nice to have a way of storing the path to a .pot file somewhere
  - And I totally agree with @vslavik about not storing the reference in the translations but maybe it should be done in the catalog manager. This way, the system _should_ work great for any file organization.

If this functionality seems interesting enough, I'll open a feature request issue.

**Longer thing about CakePHP and its translation system**
To understand why CakePHP behave like this, we must be aware of how it works. CakePHP have support for _domain strings_, which allows developpers to split their translations between functionalities ('core' domain for CakePHP core strings, then you can have as many domain as you need, creating them in the sources.). With this system, wa can use multiple translations of the same string between different sections of the application (ie, string 'Posts' for domain admin can be translated 'Posts' for another domain, and 'Articles' in another one)

When CakePHP extracts the strings from the sources to create the .pot files, the relative path is well kept, so there is no issue with that in Poedit. The weird part is the following structure:

```txt
Locale/
  + <domain>.pot files (extracted from sources, may be language agnostic)
  + fr_FR/
     + <domain>.mo files
     + <domain>.po files
```

It's done this way because of the number of possible domains and the fact that cake looks in the `Locale/<language>` folder to load the translations.

vslavik:
> I'll open a feature request issue.

Don’t open any duplicates, please. You’re free to open a *pull request* actually implementing it, though.

FWIW, I don’t recommend doing it in the manager: it’s an unmaintained, deprecated piece of crap. There’s really no reason why this shouldn’t be autodetected smartly as per my previous comment (including a CakePHP special case, because it’s no trouble for Poedit to look one level up for a POT if it can’t find it alongside the translation file).

> Longer thing about CakePHP and its translation system

None of this has bearing on this issue and is run-of-the-mill standard use of domains (modulo possible abuse of them when the same source tree — i.e. a *domain* — is artificially split into multiple gettext domains) and doesn’t really explain the peculiar choices (both standard Unix approach and WordPress’ one are domains-aware, yet don’t need to do this).

> With this system, wa can use multiple translations of the same string between different sections of the application

(BTW, I recommend reading up on “message contexts”, a long-time standard gettext feature that is actually *meant* for exactly this particular purpose…)

timofonic:
@joostdekeijzer  @mtancoigne  I see similar situations. I understand sometimes aren't done in right way, but sometimes we must live with it or there are a rational reason behind.

@vslavik Any news about this? I understand you may be busy and have other priorities. I'm just curious this 2014 pull request hasn't been solved yet.

vslavik:
@timofonic Please refrain from drive-by bugging and "me too" comments in issues. It's inappropriate and not only it doesn't help with anything, dealing with it wastes time that could otherwise be spent improving Poedit. 

The reason this PR is still open is because of its relative importance. The reason why *this form* wasn't merge is discussed at some length above, in the comments you didn't read, and is plainly declared in its very own *title* as well.

PR #141, opened 2015-01-27
Add generic undo/redo support

alexhenrie:
I don't expect this code to be accepted on the first try. Please take a good look and tell me how it could be improved.

### DIFF
--- a/src/resources/menus.xrc
+++ b/src/resources/menus.xrc
@@ -64,9 +64,9 @@
     </object>
     <object class="wxMenu" name="menu_edit">
       <label>_Edit</label>
-      <object class="wxMenuItem" name="wxID_UNDO" platform="win"/>
-      <object class="wxMenuItem" name="wxID_REDO" platform="win"/>
-      <object class="separator" platform="win"/>
+      <object class="wxMenuItem" name="wxID_UNDO" platform="unix|win"/>
+      <object class="wxMenuItem" name="wxID_REDO" platform="unix|win"/>
+      <object class="separator" platform="unix|win"/>
       <object class="wxMenuItem" name="wxID_CUT"/>
       <object class="wxMenuItem" name="wxID_COPY"/>
       <object class="wxMenuItem" name="wxID_PASTE"/>
--- a/src/text_control.cpp
+++ b/src/text_control.cpp
@@ -143,12 +143,20 @@ private:
     ITextDocumentPtr m_doc;
 };
 
-#else
+#elif defined(__WXGTK__)
 
 class UndoGroup
 {
 public:
-    UndoGroup(TranslationTextCtrl*) {}
+    UndoGroup(TranslationTextCtrl *ctrl) : m_ctrl(ctrl) {}
+
+    ~UndoGroup()
+    {
+        m_ctrl->SaveSnapshot();
+    }
+
+private:
+    TranslationTextCtrl *m_ctrl;
 };
 
 #endif
@@ -221,16 +229,21 @@ CustomizedTextCtrl::CustomizedTextCtrl(wxWindow *parent, wxWindowID winid, long 
     padding.SetRightIndent(5);
     SetDefaultStyle(padding);
 
-#ifdef __WXMSW__
+#if defined(__WXMSW__) || defined(__WXGTK__)
     Bind(wxEVT_TEXT_COPY, &CustomizedTextCtrl::OnCopy, this);
     Bind(wxEVT_TEXT_CUT, &CustomizedTextCtrl::OnCut, this);
     Bind(wxEVT_TEXT_PASTE, &CustomizedTextCtrl::OnPaste, this);
+#endif
+
+#ifdef __WXGTK__
+    if (!(style & wxTE_READONLY))
+        Bind(wxEVT_TEXT, &CustomizedTextCtrl::OnText, this);
 #endif
 }
 
 #endif // !__WXOSX__
 
-#ifdef __WXMSW__
+#if defined(__WXMSW__) || defined(__WXGTK__)
 // We use wxTE_RICH2 style, which allows for pasting rich-formatted
 // text into the control. We want to allow only plain text (all the
 // formatting done is Poedit's syntax highlighting), so we need to
@@ -239,6 +252,9 @@ CustomizedTextCtrl::CustomizedTextCtrl(wxWindow *parent, wxWindowID winid, long 
 // data when copying from the same text control to itself after its
 // content was programatically changed:
 // https://sourceforge.net/tracker/index.php?func=detail&aid=1910234&group_id=27043&atid=389153
+
+// Note that GTK has a very similar problem with pasting rich text,
+// which is why this code is enabled for GTK too.
 
 bool CustomizedTextCtrl::DoCopy()
 {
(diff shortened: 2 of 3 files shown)


vslavik:
> I don't expect this code to be accepted on the first try. Please take a good look and tell me how it could be improved.

Thanks for understanding the purpose of code review ;) It's surprisingly rare these days, I appreciate the sentiment.

I like the code overall and am pleasantly surprised how simple it was (or at least looks). I do have some nitpicking comments and some more serious ones, though:

On the nitpicking side, a more descriptive commit message would be nice, explaining why this change is done in the first place etc. Probably also worth telling it as it is: while generic, this really only has any use for GTK+. 

The only serious technical issue I can see is that I *think* it doesn't work correctly with *programmatically-set* text content. That's something `UndoSupressor` does for wxMSW (other platforms don't need or didn't so far). There are two ways text may be changed in a text control:
 1. The user does it, either by typing (`EVT_TEXT` handler) or by e.g. using a suggestion (`SetValueUserWritten()` call). This should add the change to the undo stack.
 2. It may also be changed by Poedit itself (`(Do)SetValue()` call… this may *also* result in `EVT_TEXT` (in theory it shouldn't, but platform differences… that's why you're seeing those `wxEventBlocker` in some places)). These changes should *not* be recorded and moreover, this should also clear the undo stack. I think this doesn't work correctly in this patch.

The behavior can be tested easily in Poedit: edit an entry; typing, cutting, pasting should be undoable and undo what you really did; make sure to use an entry that contains e.g. `\n` or `\"` in it so that syntax highlighting is triggered, check if undo works around the highlighted part. Using suggestions or Edit->{Clear,Copy from source} should also be undoable. Then move to another item in the list and see if you can undo then. You shouldn't, otherwise previous entry would be restored. You shouldn't be able to *redo* here either.

I'm not 100% sure this code has any problems with any of the above (I was, repeatedly, but more careful reading revealed, also repeatedly, that I was wrong) — as I said, the patch looks very good overall. I'm writing it down mainly for the benefit of testing all cases, both by you and me.

I'll add some per-line comments a little bit later.

alexhenrie:
> Wouldn't it be better to not call it in the first place in that case? Then "SaveSnapshot" would actually do it, instead of "save snapshot unless some condition". It would also be more efficient to not call OnText in the first place, i.e. do the bind on line 237 only for non-readonly controls.

SetWindowStyleFlag can add or remove wxTE_READONLY at any time after construction, so I'd rather do the check every time.

I think I've addressed all of your other feedback, including fixing a problem where the text had no highlighting after calling Undo or Redo. There is an odd problem with Paste creating a snapshot for each line as it is added instead of one snapshot for the entire paste operation, but it's not any worse of a problem than Undo/Redo only changing one user-input character at a time.

alexhenrie:
Honestly, if I were you I would switch from wxWidgets to QT, or at least use wxWidgets as a frontend to QT instead of GTK. QTextEdit and QLineEdit have fully functional undo/redo implementations on all platforms, and you'd probably be able to get rid of a lot of your other hacks too. But I know that switching toolkits would be a big, time-consuming project, and all I really need is some very basic undo/redo functionality.

vslavik:
> SetWindowStyleFlag can add or remove wxTE_READONLY at any time after construction

In theory, yes. In practice, and with *these* classes, it's *never* the case. 

> There is an odd problem with Paste creating a snapshot for each line as it is added instead of one snapshot for the entire paste operation, but it's not any worse of a problem than Undo/Redo only changing one user-input character at a time.

I disagree. The problem with one character at a time can be addressed in the same way for both affected platforms (Windows, GTK+) in the case way. It's also clear why it happens. OTOH, I have no idea why would *pasting* behave like this and it's GTK+-specific and so will have to be the fix, so it would seem better to not commit the known bug in the first place. Hmm, isn't it perhaps on syntax highlighting boundaries rather than lines (because each line will end with syntax-highlighted `\n`)? Also, it sounds like something a well-placed `UndoGroup` instance would take care of (if it prevented `SaveSnapshot()` from saving changes, which I now notice it doesn't), doesn't it?

vslavik:
> Honestly, if I were you I would switch from wxWidgets to QT

Since we're being honest: that's crazy talk. I have my share of complains and dissatisfactions with wx, oh I do, but they tend to be on the subject of wx not *being native enough*. That's a territory that Qt doesn't even begin to approach *at all*, let alone at least competently, with no hope of it ever being good (which wx is) or excellent (which it isn't) in that regard. To do UIs well, the best thing is to use the (multiple) native toolkits directly, the next best thing is wx — and then there's a huge gap with Qt somewhere far in the distance beyond it. &lt;/rant&gt;

As for using Qt port of wx on Unix instead of the GTK+ one, you only say that because you're not familiar with wxQt's freshness and unavoidable immaturity. But there's very little GTK+ specific code in Poedit, so adding support for such alternative build wouldn't be too much work.

alexhenrie:
I figured out the problem: wxWidgets on GTK copies rich text to the clipboard, then pastes it back, and when the text is pasted back it's added piece by piece. The MSW workaround works excellently for GTK too, so I just enabled the code for both platforms.

> Since we're being honest: that's crazy talk. I have my share of complains and dissatisfactions with wx, oh I do, but they tend to be on the subject of wx *not being native enough*.

QT, like wxWidgets, provides access to the underlying widget handle ([QWindow::winId](http://doc.qt.io/qt-5/qwindow.html#winId)) and has helper modules for each platform ([Qt Windows Extras](http://doc.qt.io/qt-5/qtwinextras-module.html), [Qt Mac Extras](http://doc.qt.io/qt-5/qtmacextras-module.html), [QX11Info](http://doc.qt.io/qt-5/qx11info.html)) so that you can do platform-specific manipulation. But you're right, QT is a much thicker wrapper than wxWidgets is, so if you're trying to make the wrapper as thin as possible I can see why QT would not be an attractive option for you. And like I said, I'm happy to work with your existing wxWidgets code.

Back to the topic at hand: Is the GTK UndoGroup implementation acceptable to you as-is, or do you want it to actually suppress SaveSnapshot? In practice it doesn't look like any suppression is necessary on GTK.

alexhenrie:
Another question: Do we really want the undo/redo implementation in CustomizedTextCtrl, or would TranslationTextCtrl be better?

vslavik:
> QT, like wxWidgets, provides access to the underlying widget handle … QT is a much thicker wrapper than wxWidgets is

That's a fundamental misunderstanding of the two toolkits. wx *is* a wrapper — a wx text control uses Windows' richedit control or OS X's NSTextView (or GtkTextView or QTextEdit in wxQt). Qt *does not*, it only uses the lowest windowing primitives. It reimplements all the controls from scratch and so can't use native functionality (e.g. native spellchecking that Poedit gets for free) — while wx *uses* native controls, Qt merely *resembles* them (it's not even "looks like", because there are subtle discrepancies; the illusion breaks down completely in use).

It's not about thickness, it's about being a good citizen of the target OSes.
&lt;/offtopic&gt;

> Back to the topic at hand: Is the GTK UndoGroup implementation acceptable to you as-is, or do you want it to actually suppress SaveSnapshot?

I think it's safer to suppress it, especially when it doesn't cost much in terms of code, but it's OK as it is, I'll merge it in a moment — thanks!

> Another question: Do we really want the undo/redo implementation in CustomizedTextCtrl, or would TranslationTextCtrl be better?

It's theoretically usable for non-translation edits (e.g. comments), so I think it's fine to keep it where it is.

alexhenrie:
Thank you!

PR #145, opened 2015-02-04
Revert "Fix pasting into selection to replace it on GTK+"

alexhenrie:
This reverts commit baa727e8dea01e63e3632eb2e55274f5b78b4877.

This special-casing is no longer necessary now that OnCut, OnCopy, and
OnPaste are implemented on GTK.

I'm not actually sure if you want to keep the special-casing or not, but I wanted to bring it to your attention. The special-casing doesn't seem to make a functional difference either way.

### DIFF
--- a/src/text_control.cpp
+++ b/src/text_control.cpp
@@ -524,13 +524,7 @@ void AnyTranslatableTextCtrl::HighlightText()
 
 #else // !__WXOSX__
 
-#ifndef __WXGTK__
-    // Freezing (and more to the point, thawing) the window from inside wxEVT_TEXT
-    // handler breaks pasting under GTK+ (selection is not replaced).
-    // See https://github.com/vslavik/poedit/issues/139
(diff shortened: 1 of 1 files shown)


GitCop:
There were the following issues with your Pull Request

 * Commit: e67527537fa7d12850aa0a9a071d8b868d0b2aa9

Guidelines are available at http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

---

vslavik:
> but I wanted to bring it to your attention

Thanks for that, much appreciated!

I thought about it and I think I slightly prefer leaving this code in, because it's better to not do something relatively expensive when it isn't necessary (and freezing/thawing apparently isn't on GTK+) and because it would be prone to breaking again if e.g. the pasting code changes (say, if it became possible to indicate to GtkTextView that only plaintext pasting should be done, as NSTextView allows).

PR #152, opened 2015-03-13
Fix Unix compilation warnings about WXDLLIMPEXP_*

c72578:
Apply this commit from master also to maint-1.7


vslavik:
Thanks, but the `maint-1.7` branch is in maintenance mode and only important commits that are necessary for further 1.7.x releases go into it. This is only cosmetic and keeping it on `master` only is fine.

PR #162, opened 2015-04-24
Add accelerator Shift+Enter to search backwards

alexhenrie:
This useful and common shortcut is found in Firefox, LibreOffice, and
multitudes of other applications.

I also added Enter to the accelerator table so that it behaves exactly
the same as Shift+Enter (without the delay or the animation on the
"Next" button). Still, I left "Next" as the "default button" so that it
gets the special default-button border.

### DIFF
--- a/src/findframe.cpp
+++ b/src/findframe.cpp
@@ -37,6 +37,11 @@
 #include <AppKit/AppKit.h>
 #endif
 
+#ifdef __WXGTK__
+#include <gtk/gtk.h>
+#include <gdk/gdkkeysyms.h>
+#endif
+
 #include "catalog.h"
 #include "text_control.h"
 #include "edframe.h"
@@ -176,6 +181,9 @@ FindFrame::FindFrame(PoeditFrame *owner,
     m_wholeWords->SetValue(wxConfig::Get()->ReadBool("whole_words", false));
 
     wxAcceleratorEntry entries[] = {
+#ifndef __WXGTK__
+        { wxACCEL_SHIFT,  WXK_RETURN, wxID_BACKWARD },
+#endif
 #ifdef __WXOSX__
         { wxACCEL_CMD,  'W', wxID_CLOSE },
 #endif
@@ -190,6 +198,12 @@ FindFrame::FindFrame(PoeditFrame *owner,
     Bind(wxEVT_BUTTON, &FindFrame::OnClose, this, wxID_CLOSE);
     Bind(wxEVT_MENU, &FindFrame::OnClose, this, wxID_CLOSE);
     Bind(wxEVT_CHECKBOX, &FindFrame::OnCheckbox, this);
+
+#ifdef __WXGTK__
+    GtkAccelGroup *accelGroup = gtk_accel_group_new();
+    gtk_window_add_accel_group(GTK_WINDOW(GetHandle()), accelGroup);
+    gtk_widget_add_accelerator(m_btnPrev->GetHandle(), "activate", accelGroup, GDK_KEY_Return, GDK_SHIFT_MASK, GTK_ACCEL_VISIBLE);
+#endif
 
     OnModeChanged();
     m_mode->Bind(wxEVT_CHOICE, [=](wxCommandEvent&){ OnModeChanged(); });


GitCop:

 * Commit: 6e41a36644c51b2402731709cf5d80845c8d17fd

---

vslavik:
> I also added Enter to the accelerator table so that it behaves exactly the same as Shift+Enter (without the delay or the animation on the "Next" button).

To be honest, I’m not convinced that was such a good idea. There’s a reason why standard behavior is desirable: it’s *standard*. I’d prefer normal behavior of both buttons.

Did you test this change anywhere other than whatever flavor of Linux and build of wx you use, i.e. OS X and Windows?

alexhenrie:
> To be honest, I’m not convinced that was such a good idea. There’s a reason why standard behavior is desirable: it’s standard. I’d prefer normal behavior of both buttons.

The behavior of the buttons themselves is the same. The only difference is that Enter doesn't touch the Next button anymore.

It sounds like you want the Shift+Enter keypress to be forwarded to the Previous button, so that the Previous button depresses and protrudes like the Next button. How do you propose implementing that? Do you want me to put a keypress event handler on every control in the dialog?

> Did you test this change anywhere other than whatever flavor of Linux and build of wx you use, i.e. OS X and Windows?

No, I only tested Arch Linux with wxgtk 3.0.2-2. Let me know if the behavior is significantly different on Windows or OS X.

alexhenrie:
By the way, https://wiki.wxwidgets.org/Catching_key_events_globally is a useful and relevant reference for this problem.

vslavik:
> The only difference

Yes, the “only” difference is that you add more hijacking of low-level key events — *completely untested on 2 of 3 platforms* to boot — without any clear benefit. I’d rather you didn’t, unless you have a valid reason to and clear reasoning why it’s safe and harmless (do observe, though, how the regression you recently reported was caused by such low-level messing around done years ago — *that* is why I’m being extra-cautious).

> It sounds like you want the Shift+Enter keypress to be forwarded to the Previous button, so that the Previous button depresses and protrudes like the Next button.

That would be the correct/ideal way to implement this, yes.

> How do you propose implementing that? Do you want me to put a keypress event handler on every control in the dialog?

No, I want to do *less* such ugliness, not more (plus, how exactly doing this would accomplish anything is unclear to me). It may very well not be possible with wxButton, but it *is* possible with native code (both `GtkButton` and `NSButton` allow assigning accelerator to a button). It may not be worth the extra work, and there would be a conflict with the existing Ctrl+P shortcut in the case, but still.

To sum it up:

1. Visual feedback from buttons when one is used via a shortcut is a *good*, not bad, UI thing to have.
2. It is very highly undesirable to remove this feedback from the most common case, i.e. forward search with Return.
3. It is undesirable not to have such feedback for Shift+Return, but that’s less of an issue (and so is consistency between the two buttons).

alexhenrie:
Vincent, I have force-pushed a new patch which uses a GtkAccelGroup to implement this feature for GTK only. Shift+Enter now provides the exact same visual feedback on the Previous button as Enter provides on the Next button. I think that this feature should be implemented on Windows and OSX too, but there's no reason to wait to commit this feature for GTK users.

alexhenrie:
Oh, and Alt+P still works too.

vslavik:
> but there's no reason to wait to commit this feature for GTK users.

Just because you can’t conceive of a reason or because you personally don’t care about it doesn’t mean there isn’t one. And there is: consistency, feature parity and support load (you know how people sometimes don’t even say what platform they use).

I’m not going to commit something that works only on a single platform when the feature makes sense on all of them and *can* be implemented on all of them, albeit possibly imperfectly. I don’t understand why you deleted all of the previous version, instead of only a) removing its degradation of Return handling and b) adding this improved GTK+-only handling of Shift+Return while keeping the less-than-ideal Shift+Return code (which I said above was the least problematic) for `#ifndef __WXGTK__`.

If you could update it to merge both v1 and v2 together (minus forward/Return accel), that would be great. If not, I’ll get around to finish it myself sometime soon-ish and will merge it then.

alexhenrie:
> I don’t understand why you deleted all of the previous version

I got the impression that you didn't like the previous version one bit, so I started over.

The latest version uses a wxWidgets accelerator for Windows and OSX, and a GTK accelerator for GTK. Windows doesn't do any animation when you press Enter, so there may not be an advantage to using the native toolkit instead of wxWidgets on Windows. On the other hand, I have no idea what OSX does (and I don't have access to a Mac), so you might want to write some native code to make OSX look pretty.

vslavik:
Thanks a *lot!*

Merged in b00a76e9 (your commit amended with correct ID instead of unused `wxID_BACKWARD`) and 91a8d2a6 (OS X support, fighting against wx's effective sabotage of `setKeyEquivalent:`).

alexhenrie:
> your commit amended with correct ID instead of unused wxID_BACKWARD

Thanks for catching that, that was a pretty big mistake. I actually meant to initialize m_btnPrev with wxID_BACKWARD, but your way works too.

> Thanks a lot!

You're welcome! Thanks for patiently working with me to come up with a solution that we can both agree on.

PR #163, opened 2015-04-28
Remove duplicate variable dt from DoFind

alexhenrie:


vslavik:
It’s redundant all right, but I think the code is easier to follow this way: I find it distracting to call the *current item* inside the loop `lastItem` — it’s not yet last. OTOH, once the loop ends, `lastItem` is explanatory of the value’s purpose.

alexhenrie:
Maybe we could just call it "dt" then, or some other name, or add explanatory comments? Duplicating the variable seems like a particularly ugly solution.

vslavik:
I don’t see anything particularly (or at all) ugly about it; I explained my reasoning above.

PR #164, opened 2015-04-30
Select the tab that contains the search result

alexhenrie:
Before you freak out over the goto, look this code over carefully. The goto keyword is appropriate in this case because 1) we need to break out of a double loop and 2) it avoids redundant if statements.

### DIFF
--- a/src/edframe.cpp
+++ b/src/edframe.cpp
@@ -2062,7 +2062,7 @@ void PoeditFrame::OnClearTranslation(wxCommandEvent&)
 void PoeditFrame::OnFind(wxCommandEvent&)
 {
     if (!m_findWindow)
-        m_findWindow = new FindFrame(this, m_list, m_catalog, m_textOrig, m_textTrans);
+        m_findWindow = new FindFrame(this, m_list, m_catalog, m_textOrig, m_textTrans, m_pluralNotebook);
 
     m_findWindow->ShowForFind();
 }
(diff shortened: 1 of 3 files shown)


vslavik:
Goto isn’t  inherently evil, but it does come with problems (such as unreadable execution flow as here) and the need for it is often indicative of larger issues with the code. As is the case here: the code was bad enough before and this addition pushed it over its limits. It would be better to move this inner loop to a function of its own, e.g. `IsTextInStrings` (returning index or -1 for no match). That would of course require moving the `ignoreXXX` preprocessing into `IsTextInString(s)` too, but the fact that the same preprocessing code is duplicated in `if (inSource)` branch below makes it rather clear that should have been done a long time ago anyway.

vslavik:
Most importantly, however, there’s serious breakage in this PR: while it improves things for items with plural forms, it breaks it for all the other items. `m_pluralNotebook` is only used and shown, as its name indicates, for items with *plural forms*. Most of any catalog’s items are singular-only, don’t show any notebook in the UI at all and only use `m_textCtrlTrans` (which this PR no longer uses). (And yes, this way of passing text controls to `FindFrame` as I did years ago, is a bloody mess that could use properly done API too :-/ …)

alexhenrie:
This branch has now been rebased onto the is-text-in-strings branch, and the serious breakage you pointed out is now fixed.

vslavik:
Thanks so much!

Also, apologies for the long delay — in my defense, I was busy with changes that affected translatable strings to allow as much time for translations as possible for the imminent 1.8 release.

PR #193, opened 2015-07-05
Set language code for Javanese to "jv".

jwilk:
The "jw" language code was published in error.
It was withdrawn in favor of "jv" in 2001.

Source: http://www.loc.gov/standards/iso639-2/php/code_changes.php


vslavik:
Merged because you're right (thanks!), but notice the filename — this code is irrelevant to anything but upgrading of ancient Poedit-created PO files (of which there will be approximately zero in Javanese).

PR #198, opened 2015-08-22
fix typo in French Mofidifier >> Modifier

teolemon:
fix typo in French Mofidifier >> Modifier


vslavik:
Thanks! I see you already posted the correction to [Crowdin](https://crowdin.net/project/poedit) right after creating this PR. That's indeed the preferred way, because translations are managed there, and the fix will be pulled into Poedit with the next translations update before the next release.

teolemon:
My pleasure. You've made PoEdit an invaluable ressource for Open Source translators like me :-)

PR #220, opened 2015-10-21
Changed layout of preferences/advanced under GTK

c72578:
The height of wxChoice has been increased for the selection of line endings and also the width of the wxSpinCtrl for the wrap at value.
Otherwise these settings are often not fully visible and can be difficult to set, depending on the environment

### DIFF
--- a/src/prefsdlg.cpp
+++ b/src/prefsdlg.cpp
@@ -898,7 +898,12 @@ public:
         auto crlfbox = new wxBoxSizer(wxHORIZONTAL);
         sizer->Add(crlfbox, wxSizerFlags().Expand().PXBorder(wxTOP));
         crlfbox->Add(new wxStaticText(this, wxID_ANY, _("Line endings:")), wxSizerFlags().Center().BORDER_WIN(wxTOP, PX(1)));
+#if defined(__WXGTK__)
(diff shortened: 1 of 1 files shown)


GitCop:

 * Commit: 8dca5cb990ea5fd9983e07ecf8176d6e642d4067

---

c72578:
Here would be screenshots showing preferences/advanced:
https://github.com/c72578/poedit/issues/1

vslavik:
(BTW, the ugly too-small font in your screenshots is also fixed in wx `master`)

c72578:
This is how it looks when using `m_wrapWidth = new wxSpinCtrl(this, wxID_ANY, "888");`

poedit_preferences_advanced_wxspinctrl_max_888

Remark: As soon as `wxSize(PX(50),-1)` is removed from the wxSpinCtrl, the window looks like shown as above. Independent of a max value `"888"` or `""`.

vslavik:
> Remark: As soon as wxSize(PX(50),-1) is removed from the wxSpinCtrl

Ah, yes, that must be the reason for the hardcoded value. I applied this change (although with a smaller width more in line with other platforms and as GTK+3-only code), but the wxChoice part should be fixed in wxGTK.

PR #221, opened 2015-10-26
Modify README in order to be formatted

MarioPerezEsteso:


vslavik:
I can only guess what could the rationale behind this possibly be (I recommended reading [this article]( http://chris.beams.io/posts/git-commit/) on how to convey intent in git commit messages), but regardless, I’m not going to apply this — it changes perfectly fine text file readable on any reasonable terminal, as text files should be, to insanely long lines that are not. That’s not “formatted”, that’s *un*formatted.

PR #228, opened 2015-12-06
Compilation instructions for Ubuntu

### LINKED ISSUE #227: Package names for compiling on Ubuntu?
Is there a definitive list of packages a user needs to install in order to compile the latest version?

This is what I have so far

`apt-get install libwxgtk3.0-dev wx-common  libicu-dev libstkspell-dev lucene++ boost-defaults`

I'm still getting `configure: error: missing Lucene++ library` when running `./configure`

Can someone let me know what magic I need to invoke to get the latest version installed?

edent:
Foolproof instructions to allow fools like me to quickly and easily install the latest version of poedit on Ubuntu.

Adapted from https://twistedlinux.wordpress.com/2015/10/23/install-poedit-1-8-6-from-source-on-xubuntu-14-04/

Fixes #227

### DIFF
--- a/README
+++ b/README
@@ -40,6 +40,71 @@ Optional dependencies:
   * Expat and Berkeley DB >= 4.7 (legacy translation memory migration)
   * C++REST SDK >= 2.5 (Crowdin integration)
 
+ Ubuntu
+--
+
+Installing on Ubuntu requires the installation of a specific set of dependencies.
+
+1. Install dependencies for building 
+```
+sudo apt-get install build-essential g++-4.8 g++ git make libboost1.54-all-dev libssl-dev cmake git libsecret-1-dev libgtkspell-dev
+```
+2. Create directory all required repositories 
+```
+cd ~; mkdir Source; cd Source
+```
+3. Download all required repositories
+```
+git clone https://github.com/Microsoft/cpprestsdk.git
+git clone https://github.com/luceneplusplus/LucenePlusPlus.git
+git clone https://github.com/CLD2Owners/cld2
+```
+4. Download poedit source code
+```
+wget https://github.com/vslavik/poedit/releases/download/v1.8.6-oss/poedit-1.8.6.tar.gz
+```
+5. Extract it
+```
+tar -zxf poedit-1.8.6.tar.gz
+```
+6. Build and install LucenePlusPlus
+```
+cd ~/Source/LucenePlusPlus; mkdir build; cd build
+CXX=g++-4.8 cmake .. -DCMAKE_BUILD_TYPE=Debug
+make
+sudo make install
+```
+7. Build and install REST SDK
+```
+cd ~/Source/cpprestsdk/Release; mkdir build; cd build
+CXX=g++-4.8 cmake .. -DCMAKE_BUILD_TYPE=Release
+make
+sudo make install
+```
+8. Install wxWidgets
+```
+sudo apt-key adv --fetch-keys http://repos.codelite.org/CodeLite.asc
+sudo apt-add-repository 'deb http://repos.codelite.org/wx3.0.2/ubuntu/ trusty universe'
+sudo apt-get update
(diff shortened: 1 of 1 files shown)


GitCop:

 * Commit: b75537bef7242b4f6d503f29360d61e866fa24dd

---

vslavik:
In addition to GitCop’s (ignored) issue, there are more serious ones too:

1. Firstly, what makes Ubuntu so special that it deserves distro- and version-specific instructions? Why doesn’t Fedora, RedHat, CentOS, Debian, Arch Linux, Mint, openSUSE or Gentoo, to name just a few, have it too? And why is Ubuntu 14.04 apparently more important than any other platforms (it goes before generic instructions for Linux, OS X and Windows)? This doesn’t make sense to me.

2. Even for Ubuntu, the instructions are version specific, will break with other versions (hardcoded versions) and include things already included in current Ubuntu, are redundant (pulling packages already provided by Ubuntu from obscure third party repos) and non-future-proof (ditto), insecure (blindly trusting gpg keys neither I nor you can vouch for), cargo culty (debug build of Lucene++) and resulting in poorly performing builds (ditto). Heck, they even hardcoded a specific *Poedit* version!

3. You admitted that you’re not the author. These instructions are sufficiently *creative* to probably merit copyright protection, so appropriating them like this, without the author’s express permission, is not OK. 

As I explained before,

edent:
My friend - look at the number of people asking for help using your wonderful software.  Just saying "RTFM" is doing you a great disservice.

1. Because it's the most popular.  But, by all means, include help for people with other distros. I don't really mind if it's in the README, the Wiki, or carved into stone tablets.
2. So help your users by creating a compilation script that works.  Every other piece of software I've compiled says "Install X, Y and Z version 1.2.3".  If you want new user to be able to use this software, you can't expect them to become masters of compilation.
3. You may be right. Wouldn't it be lovely if *you* could create some fool proof instructions to help spread your software far and wide?

I enjoy using your software, I want others to be able to without having to wade through complicated documents, arcane commands, and ignored issues.

Please, let me give you this little Xmas present?

Lots of love,

T

vslavik:
> My friend - look at the number of people asking for help using your wonderful software. Just saying "RTFM" is doing you a great disservice.

Kindly cut that condescending tone here. This supposed number of people who need this is tiny compared to the number of people actually using the distro-provided packages (which, IMNSHO you should too) or compiling from sources without fuss.

For the absolute majority of users on Linux, the instructions are right there at the top: “Binary easily-installable packages are provided for Windows and OS X and are part of many Linux distributions.” 

> So help your users by creating a compilation script that works. 

Poedit’s `configure` does work, checks all dependencies and ensures requires are meant. It works *exactly the same as everybody else does it*, because Autotools are the *standard* way of doing this. 

> Every other piece of software I've compiled says "Install X, Y and Z version 1.2.3”.

The README does, naturally, list the dependencies (better yet, `configure` *verifies* them!). It doesn’t enforce specific versions, because they would be magnificently stupid, but it does check and specify the version requirements that do exist.

> You may be right. Wouldn't it be lovely if you could create some fool proof instructions to help spread your software far and wide?

The instructions are already there in the README and Poedit had spread far and wide, thank you very much: it’s been included in every major distro out there (with the notable exception of server-oriented RedHat subset of Fedora) for many years.

> wade through complicated documents, arcane commands, and ignored issues.

The README is very simple. So is Ubuntu handbook’s chapter on compiling from sources. `configure` and `make` are well-known *standards*, about as far from “arcane” as you can get. You’re looking for blame everywhere except in yourself (i.e. not even reading the short chapter on how to compile on Ubuntu in the most obvious place for it to exist: in *Ubuntu* docs) and it’s not fair. Finally, there are no “ignored issues” here, that’s downright insulting from you.

> Please, let me give you this little Xmas present?

I can’t merge this PR, because it’s a likely copyright violation and of poor quality, as explained above. This shouldn’t be that hard to understand. It’s a matter of merit, emotional appeals have no relevance to that.

PR #239, opened 2016-02-14
Fix Unix installation instructions, fixes #238

### LINKED ISSUE #238: Missing configure script
According to the installation section of the [README](https://github.com/vslavik/poedit/blob/master/README):

```
 Unix
------
Do the usual thing:

    ./configure
    make
    make install
```

There should be a configure script in the root directory of the source, but no such file is found.

BTW, the attempt to generate the configure script using `autoconf` was failed with:

```
configure.ac:7: error: possibly undefined macro: AC_SUBST
      If this token and others are legitimate, please use m4_pattern_allow.
      See the Autoconf documentation.
configure.ac:15: error: possibly undefined macro: AM_INIT_AUTOMAKE
configure.ac:16: error: possibly undefined macro: AM_MAINTAINER_MODE
configure.ac:108: error: possibly undefined macro: AC_MSG_ERROR
configure.ac:110: error: possibly undefined macro: AM_CONDITIONAL
configure.ac:225: error: possibly undefined macro: AC_DEFINE_UNQUOTED
```

Vdragon:
Adding missing `./bootstrap` command before running `./configure`,
whether the instruction is still "usual" or not is left intact.

Signed-off-by: Ｖ字龍 <Vdragon.Taiwan@gmail.com>

### DIFF
--- a/README
+++ b/README
@@ -46,6 +46,7 @@ Optional dependencies:
 
 Do the usual thing:
 
+    ./bootstrap
     ./configure
     make
     make install


vslavik:
The very fact that this would not be "usual" should have tipped you off. This is not only completely unnecessary and wrong advice, but *also* described in detail further down the README for the only case when it is needed (compilation from git).

Vdragon:
Thanks for pointing out, I've stopped reading the README further once I found the "Installing from sources/Unix" section.

I would suggest to somehow improve the README by specifying those subsections only applies to released source tarballs, though.

vslavik:
> I would suggest to somehow improve the README

No, sorry, there is absolutely nothing wrong with the README, it explains everything. The section *does* apply to git checkouts, but *another* section also applies. No amount of "somehow improving" the README will help when people ignore the file's content in the first place! 

No, the *only* problem here is this:
> I've stopped reading the README

That you did that, and that you didn't return to finishing it as soon as you hit problems, is something I just don't understand.

Issue #243, opened 2016-02-22
Error while compiling in linux


pdelgadohurtado:
Error in compiling in linux manjaro (archlinux based) source from poedit-1.8.7.tar.gz

### Configure log:

`checking build system type... x86_64-unknown-linux-gnu
checking host system type... x86_64-unknown-linux-gnu
checking for a BSD-compatible install... /usr/bin/install -c
checking whether build environment is sane... yes
checking for a thread-safe mkdir -p... /usr/bin/mkdir -p
checking for gawk... gawk
checking whether make sets $(MAKE)... yes
checking whether to enable maintainer-specific portions of Makefiles... no
checking whether make supports nested variables... yes
checking for install location... /usr/local
checking for gawk... (cached) gawk
checking whether ln -s works... yes
checking for ranlib... ranlib
checking for gcc... gcc
checking whether the C compiler works... yes
checking for C compiler default output file name... a.out
checking for suffix of executables... 
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether we are using the GNU C compiler... yes
checking whether gcc accepts -g... yes
checking for gcc option to accept ISO C89... none needed
checking for style of include used by make... GNU
checking dependency style of gcc... gcc3
checking for g++... g++
checking whether we are using the GNU C++ compiler... yes
checking whether g++ accepts -g... yes
checking dependency style of g++... gcc3
checking how to run the C++ preprocessor... g++ -E
checking whether g++ supports C++11 features by default... no
checking whether g++ supports C++11 features with -std=gnu++11... yes
checking for grep that handles long lines and -e... /usr/bin/grep
checking for egrep... /usr/bin/grep -E
checking for ANSI C header files... yes
checking for sys/types.h... yes
checking for sys/stat.h... yes
checking for stdlib.h... yes
checking for string.h... yes
checking for memory.h... yes
checking for strings.h... yes
checking for inttypes.h... yes
checking for stdint.h... yes
checking for unistd.h... yes
checking cpprest/http_client.h usability... no
checking cpprest/http_client.h presence... no
checking for cpprest/http_client.h... no
checking for pkg-config... /usr/bin/pkg-config
checking pkg-config is at least version 0.9.0... yes
checking for wx-config... /usr/local/bin/wx-config
checking for wxWidgets version >= 3.0.1 (--unicode)... yes (version 3.0.2)
checking for wxWidgets static library... no
checking if wxWidgets includes XRC... yes
[...]

PR #244, opened 2016-02-23
add pthread to cpprest builds

LocutusOfBorg:
For some reasons also pthread is used on arm* boost linkages.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -73,7 +73,7 @@ AS_IF([test "x$with_cpprest" != "xno"],
               [
                   AC_MSG_CHECKING([for libcpprest >= 2.5])
                   old_LIBS="$LIBS"
-                  LIBS="-lcpprest -lboost_system -lssl -lcrypto $LIBS"
+                  LIBS="-lcpprest -lboost_system -lssl -lcrypto -lpthread $LIBS"
                   AC_LINK_IFELSE([AC_LANG_PROGRAM(
                       [
                           #include <cpprest/version.h>
@@ -96,7 +96,7 @@ AS_IF([test "x$with_cpprest" != "xno"],
 AS_IF([test "x$have_cpprest" = "xyes"],
       [
           AC_DEFINE([HAVE_HTTP_CLIENT])
-          CPPREST_LIBS="-lcpprest -lboost_system -lssl -lcrypto"
+          CPPREST_LIBS="-lcpprest -lboost_system -lssl -lcrypto -lpthread"
           AC_SUBST(CPPREST_LIBS)
           PKG_CHECK_MODULES([LIBSECRET], [libsecret-1], [
               CXXFLAGS="$CXXFLAGS $LIBSECRET_CFLAGS"


vslavik:
`-lpthread` is *never* correct. You *must* use compiled flags such as `-pthread`. These flags differ between OSes and compilers and [are hard to get right](http://www.gnu.org/software/autoconf-archive/ax_pthread.html). 

They are also included in Poedit's compilation courtesy of wx, so the second part is certainly unnecessary; for the former, I'm beginning something is seriously broken somewhere else.

I'll be adding Boost configure checks for #243, the proper fix will then be to pass various Boost flags to C++REST test (incl. correct flags for Boost.Thread, which C++REST uses).

LocutusOfBorg:
I don't know why, and I know too little autoconf to provide a useful patch, but I did try to understand why the build wasn't taking cpprest enabled on arm64 and armhf, even if libcpprest-dev was installed
https://buildd.debian.org/status/fetch.php?pkg=poedit&arch=arm64&ver=1.8.7-2&stamp=1456231117
and when I tried to link an example code the reason was missing -lpthread.
So, I think I'll apply this patch as Debian-specific patch for now, and if you have a better solution I'll be glad to apply that one too :)

vslavik:
> So, I think I'll apply this patch as Debian-specific patch for now

Please, don't be that kind of maintainer 😞 The reason gcc has the `-pthread` flags is *so that you use it*. So, please, use it. As long as you're Linux+GCC/Clang only, it's fine. I can't merge it, because it's broken basically anywhere else, but it's fine *for you*. `-lpthread` isn't.

LocutusOfBorg:
"Please, don't be that kind of maintainer :disappointed: The reason gcc has the -pthread flags is so that you use it. So, please, use it. As long as you're Linux+GCC/Clang only, it's fine. I can't merge it, because it's broken basically anywhere else, but it's fine for you. -lpthread isn't. "

oh... it was a compiler flag, not linker flag.
Agreed, I'll change it then :) sorry for missing that point

LocutusOfBorg:
this is the new patch
```
--- poedit-1.8.7.1.orig/configure.ac
+++ poedit-1.8.7.1/configure.ac
@@ -68,7 +68,7 @@ AS_IF([test "x$with_cpprest" != "xno"],
       dnl it for this header too and it uses only the preprocessor in one
       dnl of its two phases:
       old_CPPFLAGS="$CPPFLAGS"
-      CPPFLAGS="$CPPFLAGS $CXXFLAGS"
+      CPPFLAGS="$CPPFLAGS $CXXFLAGS -pthread"
           AC_CHECK_HEADERS([cpprest/http_client.h],
               [
                   AC_MSG_CHECKING([for libcpprest >= 2.5])
@@ -99,7 +99,7 @@ AS_IF([test "x$have_cpprest" = "xyes"],
           CPPREST_LIBS="-lcpprest -lboost_system -lssl -lcrypto"
           AC_SUBST(CPPREST_LIBS)
           PKG_CHECK_MODULES([LIBSECRET], [libsecret-1], [
-              CXXFLAGS="$CXXFLAGS $LIBSECRET_CFLAGS"
+              CXXFLAGS="$CXXFLAGS $LIBSECRET_CFLAGS -pthread"
               AC_SUBST(LIBSECRET_LIBS)
           ])
       ],
```

vslavik:
I believe 0314d61f addresses this.

Issue #262, opened 2016-03-31
Assertion failure with wxEXPAND and wxBOTTOM


Flamefire:
In version 3.1 of wxWidgets using both of those options results in an assertion failure. This is reproducible by opening a pot file:

    wxWidgets-3.1.0/src/common/sizer.cpp(2116): assert "!(flags & (wxALIGN_BOTTOM | wxALIGN_CENTRE_VERTICAL))" failed in DoInsert(): Vertical alignment flags are ignored with wxEXPAND

One of those is here: https://github.com/vslavik/poedit/blob/c09fcac53ea5bdc2aa10e54c54753abe62fb1fdb/src/resources/properties.xrc#L30

vslavik:
wx 3.1 is *development* series and Poedit isn’t meant to be compiled with it. There are far more issues with it than this (as it happens, I’m *just* working on it). By all means, use devel series, but if you do, don’t file issues — file *PRs*.

> One of those is here:

Nope, that’s a completely wrong guess by you, the actual issue was `wxALIGN_CENTRE_VERTICAL` elsewhere in the file.

Flamefire:
According to https://www.wxwidgets.org/ 3.1.0 is actually a release and Poedit just says >=3.0. So I assumed that this should work. 

And nope,  `wxALIGN_CENTRE_VERTICAL` is nowhere used in that file. However the real issue was in some cpp files. I opened a PR for this.

vslavik:
Thank you for your lecture. [I only contributed to it for ~10 years](https://www.wxwidgets.org/about/credits/) so you clearly have better understanding of what they do. wx 3.1 is a *development* release and is clearly marked as such on the website, as I previously tried to explain to you.

> And nope, wxALIGN_CENTRE_VERTICAL is nowhere used in that file.

[Ahem…](
https://github.com/vslavik/poedit/pull/263/commits/1e40cb7e0de16f8e08ae056a99dd0d24e582081c#diff-0dd281efedae63f7cca7b449bcaeae51L237)

graggi:
At the end, did you solve this problem? and if you did how can I correct it?

Thanks

PR #263, opened 2016-04-01
Fix assertion failures with wxWidgets 3.1

Flamefire:
As described in #262 there are a couple of assertion failures with the new wxWidgets version due to the use of ` Expand()`  with alignment flags.
In ` crowdin_gui.cpp` it is illegal to use `Right` in a horizontal sizer
And last `wxALIGN_CENTER_VERTICAL` in the properties.xrc is invalid with this flags or at this point. Suggestion by the assert was to use spacers. Doesn't look bad without them however


vslavik:
Thanks, but I’m not going to merge this — as I mentioned in the other bug, I fixed this particular thing already, just didn’t push it yet. More importantly, your 1e40cb7 breaks preferences layout in at least two places.

PR #288, opened 2016-09-06
avoid deprecated gtk_widget_size_request()

c72578:
gtk_widget_size_request has been deprecated since GtkWidget version 3.0.

Avoids compiler warnings like:

`wx_backports/wx_gtk_nativewin.cpp: In member function ‘bool xNativeWindow::Create(wxWindow*, wxWindowID, wxNativeWindowHandle)’:
wx_backports/wx_gtk_nativewin.cpp:64:5: warning: ‘void gtk_widget_size_request(GtkWidget*, GtkRequisition*)’ is deprecated: Use 'gtk_widget_get_preferred_size' instead [-Wdeprecated-declarations] gtk_widget_size_request(widget, &req);`

### DIFF
--- a/src/wx_backports/wx_gtk_nativewin.cpp
+++ b/src/wx_backports/wx_gtk_nativewin.cpp
@@ -61,7 +61,7 @@ wxNativeWindow::Create(wxWindow* parent,
     // Ensure that the best (and minimal) size is set to fully display the
     // widget.
     GtkRequisition req;
-    gtk_widget_size_request(widget, &req);
+    gtk_widget_get_preferred_size(widget, NULL, &req);
     SetInitialSize(wxSize(req.width, req.height));
 
     return true;
--- a/src/wx_backports/wx_gtk_nativewin.cpp
+++ b/src/wx_backports/wx_gtk_nativewin.cpp
@@ -61,7 +61,11 @@ wxNativeWindow::Create(wxWindow* parent,
     // Ensure that the best (and minimal) size is set to fully display the
     // widget.
     GtkRequisition req;
+#ifdef __WXGTK3__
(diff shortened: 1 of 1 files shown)


c72578:
Travis CI build failed.
Distinction of cases for GTK+ 3 and GTK+ 2 required, will work on this ...

vslavik:
Quoting Travis build:

    ../../src/wx_backports/wx_gtk_nativewin.cpp: In member function ‘bool wxNativeWindow::Create(wxWindow*, wxWindowID, wxNativeWindowHandle)’:
    ../../src/wx_backports/wx_gtk_nativewin.cpp:64:53: error: ‘gtk_widget_get_preferred_size’ was not declared in this scope
         gtk_widget_get_preferred_size(widget, NULL, &req);

Per the documentation, the function is GTK+ 3.0 only. wx has [compat shims](https://github.com/wxWidgets/wxWidgets/blob/d96a25be1230b91f3431bfc3b85b4e1f55bf62e6/include/wx/gtk/private/gtk2-compat.h) to deal with this that are private, i.e. unavailable for Poedit code. So you either have to use `#ifdefs` or live with the warning. As the filename indicates, this is backported wx code that isn’t used when compiling with wxGTK 3.1 (which I strongly recommend if using GTK+ 3, as you seem to be), so the latter doesn’t seem unreasonable to me...

c72578:
Added `#ifdef __WXGTK3__`

c72578:
Here is some additional information, why to get rid of the compiler warning.
In Fedora Linux we have wxGTK3 version 3.0.2 at the moment. So, this is what Poedit is compiled with when creating the rpms.
https://apps.fedoraproject.org/packages/wxGTK3

Therefore not wxGTK 3.1 and my PR instead.

PR #354, opened 2017-03-19
Plural expression for ie

Carmina16:
Adding support for the Interlingue plural expression.


vslavik:
Please follow [best git commit message practices](https://chris.beams.io/posts/git-commit/) — “Interlingue plural is not a descriptive commit message". Please don’t add duplicate commits that don’t even do what their subject says. 

Finally, please also observe that the file you modified says it was generated by `scripts/extract-plural-forms.py` — in fact, it says so on a line you modified. You have to make changes to that script, not the generated file, for them to have a lasting effect.

vslavik:
In the future please *amend* your PRs (both git and GitHub has ample documentation) instead of making a gigantic mess by closing PRs with no explanation and replacing them with the same thing anew, thus loosing any history.

Replaced by #356, apparently.

Issue #356, opened 2017-03-23
Opening a po file at a given segment number


pepc:
It would be great if poedit had a command-line argument that allowed to open a .po file (or reposition a currently opened .po file) to a given segment, something like `poedit.exe myfile.po -segment=10` to put the segment cursor on the the 10th segment in the file.

This would allow for example a third-party QA tool to launch poedit right at the segment, so that it can be edited in poedit.

vslavik:
Are you jus theorizing or do *you*, personally, have actual, present need for this? Will anything actually use such function if it is implemented? If you do, please provide more specific details. 

By the way, “segment” is ill-suited terminology, it’s never used in gettext. And most lint-kind tools report line numbers with problems, so opening at an item corresponding to *a line in the file* would make more sense IMHO.

pepc:
It was a request of one of our customers for a [QA tool](https://www.xbench.net).

In our QA tool, Xbench, we load and parse bilingual files but we do not edit them.  Instead, we call the home editor (Trados Studio, memoQ, Memsource, etc) and let the user edit the segment there.  This workflow avoids data integrity issues (some of these formats are proprietary so there are risks by editing files even if we parse them fine for browsing).  Around minute 2:20 in [this video](https://www.youtube.com/watch?v=TDEM1oXH8D4) you can see how it works with the Memsource Editor.  In our [YouTube Channel](https://www.youtube.com/user/xbenchdotnet) there are other similar integrations illustrated.

The instant gratification of seeing an issue in Xbench and instantly fix it in the home application with a keystroke is highly appreciated by users. 

Line number as input would be also fine, (i.e. something like `poedit.exe myfile.po -atline=123` because our parser can store line numbers in our internal data structures and then call poedit with the line number as argument.

The main audience for this feature would probably be professional translators who are very conscious of linguistic quality and use our tool (for example, they are translating with poedit and want to instantly spellcheck all .po files.

vslavik:
I see. So since your interest in this is commercial, perhaps you’d like to make a PR? This shouldn’t be too difficult and I’m happy to lend a hand… but to be honest, I’m too busy finishing v2.0 at the moment to work on this myself (although I do agree it should be done).

pepc:
Sure, we can give it a try.  In fact, yesterday we tried twice to clone from Github using Sourcetree but the clone failed at a couple of dependencies.  Will try to clone again, perhaps using the Github desktop client instead.

vslavik:
[...]

PR #358, opened 2017-03-25
Support for command-line switch --line

pepc:
This adds support for the new command-line switch --go-to-line=<num>,
where <num> is the number of line in the .po file. This is related to
issue #356 on vslavik/poedit. It has been compiled with Visual Studio 2015 and tested on Windows.

### DIFF
--- a/src/catalog.cpp
+++ b/src/catalog.cpp
@@ -1365,13 +1365,19 @@ void Catalog::RemoveDeletedItems()
 
 CatalogItemPtr Catalog::FindItemByLine(int lineno)
 {
-    CatalogItemPtr last;
+    int i = FindItemIndexByLine(lineno);
+    return i == -1 ? CatalogItemPtr() : m_items[i];
+}
+
+int Catalog::FindItemIndexByLine(int lineno)
+{
+    int last = -1;
 
     for (auto& i: m_items)
     {
-        if ( i->GetLineNumber() > lineno )
+        if (i->GetLineNumber() > lineno)
             return last;
-        last = i;
+        last++;
     }
 
     return last;
--- a/src/catalog.h
+++ b/src/catalog.h
@@ -709,6 +709,9 @@ class Catalog
         /// Finds item by line number
         CatalogItemPtr FindItemByLine(int lineno);
 
+        /// Finds catalog index by line number
+        int FindItemIndexByLine(int lineno);
+
         /// Sets the given item to have the given bookmark and returns the index
         /// of the item that previously had this bookmark (or -1)
         int SetBookmark(int id, Bookmark bookmark);
--- a/src/edapp.cpp
+++ b/src/edapp.cpp
@@ -158,9 +158,11 @@ private:
             }
             if (data.StartsWith("OpenFile:", &payload))
             {
+                long lineno = 0;
+                payload.BeforeFirst(':').ToLong(&lineno);
                 wxArrayString a;
-                a.push_back(payload);
-                m_app->OpenFiles(a);
+                a.push_back(payload.AfterFirst(':'));
+                m_app->OpenFiles(a, lineno);
                 return true;
             }
             return false;
@@ -234,11 +236,11 @@ public:
         Command("OpenURI:" + uri);
     }
 
-    void OpenFile(const wxString& filename)
+    void OpenFile(const wxString& filename, int lineno = 0)
     {
         wxFileName fn(filename);
         fn.MakeAbsolute();
-        Command("OpenFile:" + fn.GetFullPath());
+        Command(wxString::Format("OpenFile:%d:%s", lineno, fn.GetFullPath()));
     }
 
 private:
@@ -312,6 +314,7 @@ bool PoeditApp::CheckForBetaUpdates() const
 #ifndef __WXOSX__
 static wxArrayString gs_filesToOpen;
 #endif
+static int gs_lineToOpen = 0;
 
 extern void InitXmlResource();
 
@@ -447,8 +450,9 @@ bool PoeditApp::OnInit()
     //     passing files on command line
     if (!gs_filesToOpen.empty())
     {
-        OpenFiles(gs_filesToOpen);
+        OpenFiles(gs_filesToOpen, gs_lineToOpen);
         gs_filesToOpen.clear();
+        gs_lineToOpen = 0;
     }
     else
     {
@@ -605,7 +609,7 @@ void PoeditApp::OpenNewFile()
         PoeditFrame::CreateWelcome();
 }
 
-void PoeditApp::OpenFiles(const wxArrayString& names)
+void PoeditApp::OpenFiles(const wxArrayString& names, int lineno)
 {
     PoeditFrame *active = PoeditFrame::UnusedActiveWindow();
 
@@ -628,12 +632,12 @@ void PoeditApp::OpenFiles(const wxArrayString& names)
 
         if (active)
         {
-            active->OpenFile(name);
+            active->OpenFile(name, lineno);
             active = nullptr;
         }
         else
         {
-            PoeditFrame::Create(name);
+            PoeditFrame::Create(name, lineno);
         }
     }
 }
@@ -664,6 +668,7 @@ namespace
 {
 const char *CL_KEEP_TEMP_FILES = "keep-temp-files";
 const char *CL_HANDLE_POEDIT_URI = "handle-poedit-uri";
+const char *CL_LINE = "line";
 }
 
 void PoeditApp::OnInitCmdLine(wxCmdLineParser& parser)
@@ -674,6 +679,8 @@ void PoeditApp::OnInitCmdLine(wxCmdLineParser& parser)
                      _("don't delete temporary files (for debugging)"));
     parser.AddLongOption(CL_HANDLE_POEDIT_URI,
                      _("handle a poedit:// URI"), wxCMD_LINE_VAL_STRING);
+    parser.AddLongOption(CL_LINE,
+                     _("go to item at given line number"), wxCMD_LINE_VAL_NUMBER);
     parser.AddParam("catalog.po", wxCMD_LINE_VAL_STRING,
                     wxCMD_LINE_PARAM_OPTIONAL | wxCMD_LINE_PARAM_MULTIPLE);
 }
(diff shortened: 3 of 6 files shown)


vslavik:
Wow, that was fast, thanks!

Can you please update it to follow Poedit’s coding convention, though, at least for the most jarring issues immediately apparent in the diff, i.e. use of tabs and use of trailing `{` (Poedit puts it on a new line and uses 4-spaces indentation)?

----

I have some more feedback, but am willing to do them myself (of course, help with them would be appreciated):

1. It seems the de facto standard way of doing this command line argument in Unix editors is to use `+[num]` (see [gedit](http://linuxcommand.org/man_pages/gedit1.html), vim is the same). You probably have more experience with how other different editors handle this, though? 

2. For the IPC communication on Windows, it’s arguably better to put both (optional) line number and filename into `OpenFile:` (as e.g. `OpenFile::foo.cpp` and `OpenFile:42:foo.cpp` respectively) because they make no sense as separate commands, the line is always bound to the file.

3. I have a feeling that the actual repositioning is a bit too complicated, but perhaps I’m just missing some issues you encountered? In any case, you don’t have to worry about race conditions here, all this code, including `CallAfter` blocks, is executed on the same thread in deterministic order.

GitCop:

 * Commit: 2e39f1e83913df585b01d124e9ac9a0010c7aa70

---

pepc:
Sorry, it seems that GitHub did not like the verbose description of my second commit in the branch (I'm a git/GitHub noob). I'll try to fix it later (I'm not familiar with `git push -f`).

In the second commit, I think I fixed all isses with tabs and curly braces.  I also clarified that it only supports positioning for the first (which is the same thing gedit does, according to the description on the landing page of your link).

In our use case, we will always place the call with exactly one file as argument, which I suppose is what will do other interactive programs who wish to go at a specific line.  Perhaps this context of interactivity explains why gedit also supports positioning for only the first file in the argument list.

However, if you wish to support positioning for more than one file, then perhaps I would suggest to follow [SublimeText convention](http://docs.sublimetext.info/en/latest/command_line/command_line.html), that is follow the file name with :line or :line:column because it is compact, and leverages the fact that the colon cannot be in a file name.

If you choose the gedit syntax, please note it is possible that you have to change your current command line parser, as I think it expects a keyword and then a value.

As for race conditions if there is only one thread then it does not make sense, but if I'm not wrong, I noticed that if a dialog box call was issued at certain places (which I did while debugging), sometimes the `SelectAndFocus(0)` at `OnCatalogChanged` could be executed before the one at `PlaceInitialFocus`.  I also liked the idea that positioning took effectively place always at `PlaceInitialFocus`.  I agree that the code looks pretty complicated, but to support the cases of file already loaded and file not loaded yet, I did not come up with anything simpler (don't know enough the code to refactor the flow of relevant events without breaking anything).

vslavik:
Sorry for the delay, releasing Poedit 2 was... hectic. Will look at this soon, for inclusion in 2.0.2.

vslavik:
I force-pushed updated version to this PR's branch: cleaned up the code, avoid ugly globals, renamed the option to `--line` (as a compromise between clarity and length), fixed some bugs (broken if non-default sorting used), addressed point 2 and partially 3 above.

Given the relatively large extent of changes, do you think you could re-test it if everything works (seems to here, but…)? TIA!

pepc:
I tried to build on VS 2015 and there seems to be an issue with the wxWidgets_wxscintilla dependency.  

wxWidgets_wxscintilla fails to build with this error:

4>c1xx : fatal error C1083: Cannot open source file: '..\wx\src\stc\scintilla\lexers\LexEDIFACT.cxx': No such file or directory

vslavik:
> I tried to build on VS 2015 and there seems to be an issue with the wxWidgets_wxscintilla dependency.

No, there's an issue with the way you checked out from git. See the README for how to update submodules.

pepc:
OK, I simply did a clone from scratch with Github Desktop, which worked well the last time.

Out of curiosity: Why is there a "continuous-integration/travis-ci/pr" fail report in this PR thread for your commit?

vslavik:
> Out of curiosity: Why is there a "continuous-integration/travis-ci/pr" fail report in this PR thread for your commit?

There's a Details link for a reason -- to be clicked if you're curious… Don't worry about it.

pepc:
Just checked and it works well with these scenarios:
a) PO file is not yet open and poedit.exe --line is launched
b) PO file is open but not yet saved and poedit.exe --line is launched

If does not work with this scenario:
a) PO file is open and has been saved poedit.exe --line is launched

This is a reproducible test case for the save issue (reproducible with file `locales\ca.po` in the poedit repo).

1. Run `poedit ca.po --line 2010`   ==> It opens file ca.po and correctly goes to segment "&Apply"
2. Click the **Save** button in Poedit
3. Run `poedit ca.po --line 2010` ==> It goes to the wrong segment (in this case "shift", at the end of the file).

If you need more detailed info on how to reproduce the issue after clicking **Save**, just let me know.

vslavik:
Thanks for discovering this bug (it was present in the original version too, BTW). I have now addressed it by reloading the file when Poedit is told to open an already-open file. This means that this case:

> b) PO file is open but not yet saved and poedit.exe --line is launched

will now show a message to the user asking if they want to save their changes, but I think that's actually correct: if you invoke Poedit asking it to *open a file* (be it at a specific line or not), it should attempt to actually open the file as it exists on disk, shouldn't it?

pepc:
From a UX standpoint, and at least with the use case of interactive QA that we have in mind, it was better user experience to not reload the file.  The reasons are these:

a) The save/load is a relatively expensive task in poedit.  Even this the ca.po file of only 50KB you may easily get DDE timeout warning messages, especially if there are "Do you want to save?" in the process.

b) Since this is interactive QA, the typical workflow involves going back and forth between the QA with Alt+Tab.  The "Do you want to save?" message each time will worsen the UX experience.

c) Normally, when translating no new segments are added.  This means that the line-to-item mapping of the 3thd party QA tool matches the po file saved.  If you save the po file in the middle of the call, if new lines are added the mapping can be broken.

The typical translator process is fix, fix, fix, fix, fix, then save, then reload in our tool for a regression to ensure nothing is left. The fix, do you want to save, fix, do you want to save, fix, do you want to save, fix, will be a lot more tedious for her.

My preferred solution would be to ensure that the internal line-to-item mapping in `FindItemIndexByLine` continues to match the file that was just saved.  If the file was saved, IMHO it is the responsibility of the 3rd party tool (or the user) to reload it.

If that approach is not trivial, another option is not to issue the "Do you want to save?" message. For example with Memsource Editor, if you try to open the same file and it is open it saves the current buffer to disk and then opens the file (we did not ask for it, it was already working like this when we implemented the integration).

If you wish, to better visualize the UX of the use case for inteactive QA, we can post a dev build of our tool with support for the `--line` argument.

vslavik:
> it was better user experience to not reload the file

…and to jump to the wrong line in some situations? The reason I made this change is two-fold: fixing #393 will not be entirely trivial and reloading files (if modified externally) seemed to be the right thing to do regardless of everything else at first glance (although you made me reconsider).

To be blunt, it's not feasible for me to fix #393 in time for 2.0.2 — and it make take even more, depending on how difficult this turns out to be. So at the moment you get to pick from two bad options:

(1)
"PO file is open but not yet saved and poedit.exe --line is launched" works
"PO file is open and has been saved poedit.exe --line is launched" may jump to a wrong item

or

(2)
"PO file is open but not yet saved and poedit.exe --line is launched" works, but bugs the user with a reload question and is slower
"PO file is open and has been saved poedit.exe --line is launched" works

Which is better? Based on your response and the fact that you didn't even notice the bug 167d668 is trying to address in your initial implementation, I take it (1) is preferable?

> Even this the ca.po file of only 50KB you may easily get DDE timeout warning messages

That's a clear bug, it should be an async operation.

> c) Normally, when translating no new segments are added. This means that the line-to-item mapping of the 3thd party QA tool matches the po file saved.

It does not mean that. Any changes to the translations may (and typically do!) affect line formatting of the output file and shift subsequent items — unless you use the nonstandard no-wrapping formatting of POs.

> My preferred solution would be to ensure that the internal line-to-item mapping in FindItemIndexByLine continues to match the file that was just saved.

Well, yes, obviously, but it's easier said that done due to some extra post-processing of the file when saved — it is passed through `msgcat` for reformatting and would have to be loaded again internally just to determine the line numbers (something that takes a bit of time, as you observed, and that won't even be needed for anything in most cases).

>  If the file was saved, IMHO it is the responsibility of the 3rd party tool (or the user) to reload it.

Arguably the user invoking `poedit.exe ca.po` is just such explicit user action to reload the file, hence 167d668... (For the record, I don't entirely agree with you: ideally, Poedit would detect that the file was changed underneath it and either reload automatically if no changes or ask the user, as e.g. Visual Studio does.)

> If that approach is not trivial, another option is not to issue the "Do you want to save?" message. For example with Memsource Editor, if you try to open the same file and it is open it saves the current buffer to disk and then opens the file 

Seems wrong to silently overwrite user's file without explicit instruction…

I think the best solution *for now* is to have the initial implementation without 167d668 in 2.0.2 (as you say, typical process is not saving until the user is done with a batch of fixes). Hopefully I'll come with an efficient fix for #393 for 2.0.3 and even if not, it shouldn't be noticeable in the most common workflow.

What do you think?

pepc:
I think I prefer option (1) (i.e. your initial commit) because if the user happens to save in poedit, she still has the workaround of closing the file in poedit (or closing poedit) and issuing the call again from Xbench. In fact I think users are going to like how fast it is to position the file when it is already open.

Once 2.0.2 is released with `--line` support, we can add an issue in Github with the save issue discussed so that it can be considered for fixing at a later stage, because at least in my opinion it is not a "ship issue".

Besides, it will be very useful to know that starting with version 2.0.2 we can count on a `--line` argument to put the funcionality in place, because we have to use poedit.exe file version info to discriminate if we can add or not a `--line` argument in our call (i.e. if we add a `--line` argument against 2.0.1 or earlier, we will get an error message).

vslavik:
> at least in my opinion it is not a "ship issue".

Agreed. The bug for tracking it is already there, btw: #393.

Merged now, thanks a lot for your help with the implementation!

PR #389, opened 2017-04-21
Add the plural expression for Interlingue

Carmina16:
New plural expression for Interlingue language.


vslavik:
Starting with 61d4394388737fb6228c668d622a04bec0426d53 Poedit uses CLDR data to generate plurals. It is not feasible to manage them manually as this PR does, so I'm closing it.

Interlingue support will be added as soon as the corresponding data appear in CLDR. As long as not even CLDR carries stuff like this, I'm fine with Poedit not doing it either — especially when you can always enter your own custom expression.

Issue #390, opened 2017-04-29
Hidden font selection button in the preference dialog


tvincent2:
Hello,

I noticed that in the preference dialog of poedit, the button to set the font for text fields is nearly invisible (see [attached capture](https://people.debian.org/~tvincent/img/poedit-preference-dialog.png)).

I can reproduce it with poedit 1.8.11 (from Debian Stretch) and 2.0.1 (from experimental). I am using Gnome Shell with a fairly default config, notably default fonts (Cantarell Regular, 11).

Please let me know if I can be of any help to reproduce and fix this bug.

Thanks a lot for your work on poedit!
Thomas

vslavik:
Please actually *attach* attachments instead of linking to a random site (where they may disappear at any time).

Please provide full reproduction instructions (see [this](http://www.chiark.greenend.org.uk/~sgtatham/bugs.html)), including such basics as wxGTK and GTK versions. 

This looks like a wxGTK bug to me, so checking with its latest version seems advisable.

tvincent2:
I am not used to issues in github, sorry. :)

Here is the attachment.

![poedit-preference-dialog](https://cloud.githubusercontent.com/assets/2185278/25555751/deab1b92-2cee-11e7-97ba-3116b99644fb.png)

tvincent2:
Hello,

My system has wxGTK in version 3.0.2+dfsg-4 and GTK+ in version 2.24.31-2, both installed from Debian Stretch.

I am not sure what other info is needed, so please let me know how I can help. :)

Thomas

LocutusOfBorg:
I have the same issue with Ubuntu Xenial, Wxwidgets 3.0.2+dfsg-1.3 gtk+ 2.12.10-6

vslavik:
Let's see if anything changes once running against wxGTK 3.0.3 (recently released, so hopefully will be packaged not too long from now too).

LocutusOfBorg:
Hello, I installed a clean machine (stretch based) and I could reproduce only with the poedit from experimental
(so, 1.8.11 was fine, 2.0.1 not)
I packaged wxGTK 3.0.3 (you can use the debs from here http://debomatic-amd64.debian.net/distribution#experimental/wxwidgets3.0/3.0.3+dfsg-1~exp1/buildlog )
Note, I didn't check for API/ABI breaks, so install it in some controlled environment :)

Testing was the same, still broken the experimental version.

c72578:
[...]

PR #411, opened 2017-07-07
Fix overlapping crowdin name and username (wxGTK3)

c72578:
- Under wxGTK3, there is not enough space allocated for the name. Longer names
  are wrapped and overlap with the username then. See screenshot below.
- Hard code an initial horizontal width for wxSize and ellipsize longer
  names as a workaround

Poedit version: 2.0.2
OS: Fedora 26 x86_64
wxGTK3 version: 3.0.3
GTK+ version: 3.22.16
 
Overlapping name and username in Poedit 2.0.2:

poedit_2 0 2_prefs_crowdin_screenshot_name_overlap

This is how it looks with the modified crowdin_gui.cpp:

poedit_2 0 2_prefs_crowdin_screenshot_name_no-overlap

### DIFF
--- a/src/crowdin_gui.cpp
+++ b/src/crowdin_gui.cpp
@@ -166,7 +166,13 @@ void CrowdinLoginPanel::CreateLoginInfoControls(State state)
         {
             auto account = new wxStaticText(this, wxID_ANY, _("Signed in as:"));
             account->SetForegroundColour(SecondaryLabel::GetTextColor());
+#ifdef __WXGTK3__
+            // Under wxGTK3, there is not enough space for the name. Longer names are wrapped and overlap with the username then
+            // Hard code an inital horizontal width for wxSize and ellipsize longer names as a workaround
+            auto name = new wxStaticText(this, wxID_ANY, m_userName, wxPoint(), wxSize(150,-1), wxST_ELLIPSIZE_END);
+#else
             auto name = new wxStaticText(this, wxID_ANY, m_userName);
+#endif
             name->SetFont(name->GetFont().Bold());
             auto username = new SecondaryLabel(this, m_userLogin);
 
@@ -174,6 +180,8 @@ void CrowdinLoginPanel::CreateLoginInfoControls(State state)
             sizer->AddSpacer(PX(2));
             auto box = new wxBoxSizer(wxVERTICAL);
             box->Add(name, wxSizerFlags().Left());
+            // A spacer in between would also improve the overlapping of name and username under wxGTK3
+            // box->AddSpacer(PX(25));
             box->Add(username, wxSizerFlags().Left());
             sizer->Add(box);
             break;


vslavik:
Thanks, but this is not a matter of there "no being enough space" — there clearly is plenty. 

I think this is just a demonstration of one of *many* GTK+3 problems when using wxGTK 3.0. I realize this is out of your hands, but this class of problems is best solved by using wxGTK 3.1 + GTK+3 or wxGTK 3.0 + GTK+2. 

Did you confirm with wxGTK 3.1 (master) by any chance? If not, I will before applying this and would apply this only for the very limited circumstance of wxGTK 3.0 w/ GTK+3. Even then, setting the size to something reasonable (i.e. derived from the actual size of the `m_userName` text) instead of magic constant that works well for "Wolfgang Stöggl" but not others would be preferable...

c72578:
Thanks for the feedback.
I have just built Poedit against wxGTK 3.1 (master, commit [2d8657d)](https://github.com/wxWidgets/wxWidgets/commit/2d8657d37cbf71701dc28047b5f4f6ebde2d8df0) and could confirm the issue.

vslavik:
Thanks! 

Just in case, you didn't notice any similar stuff elsewhere in Poedit, did you (#390 is something else)?

c72578:
I did not notice anything else.
Except the already known https://github.com/vslavik/poedit/pull/220, where the layout is at least "usable" although it does not look as intended.

c72578:
Here is something else I noticed in the meantime, concerning the sidebar with translation suggestions and the notes for translators, which you are probably aware of (e.g. https://github.com/vslavik/poedit/commit/5aaf000abd058bf3d35541451c5df4f58127c632)
Overlapping text:
Translation suggestions:
Notes for translators:

The "Notes for translators:" is above parts of the notes, which makes it difficult to read the notes.

poedit_master_sidebar_overlap_2017-07-13

Looks the same with wxGTK 3.0.3 or wxGTK 3.1 (master).

vslavik:
In other words, wx's `wxStaticText` sizing is still totally broken with GTK+ 3 :( May be this (supposed fixed) wx bug: http://trac.wxwidgets.org/ticket/17585

vslavik:
This is an alternative, if unfortunate, workaround:

```diff
diff --git a/src/crowdin_gui.cpp b/src/crowdin_gui.cpp
index a6dd69d534..ff0d158439 100644
--- a/src/crowdin_gui.cpp
+++ b/src/crowdin_gui.cpp
@@ -167,7 +167,7 @@ void CrowdinLoginPanel::CreateLoginInfoControls(State state)
             auto account = new wxStaticText(this, wxID_ANY, _("Signed in as:"));
             account->SetForegroundColour(SecondaryLabel::GetTextColor());
             auto name = new wxStaticText(this, wxID_ANY, m_userName);
-            name->SetFont(name->GetFont().Bold());
+            //name->SetFont(name->GetFont().Bold());
             auto username = new SecondaryLabel(this, m_userLogin);
 
             sizer->Add(account, wxSizerFlags().BORDER_MACOS(wxTOP, PX(3)));
diff --git a/src/customcontrols.h b/src/customcontrols.h
index c90135755c..6765353a09 100644
--- a/src/customcontrols.h
+++ b/src/customcontrols.h
@@ -46,7 +46,7 @@ public:
     HeadingLabel(wxWindow *parent, const wxString& label)
         : wxStaticText(parent, wxID_ANY, label)
     {
-        SetFont(GetFont().Bold());
+        //SetFont(GetFont().Bold());
     }
 
 };
```

That pretty much confirms it's wx bug #17585 I think.

vslavik:
Actually, it's this (still open) wx bug: https://trac.wxwidgets.org/ticket/16088

vslavik:
The above commit uses a different approach to sidestep the wx bug by using Pango markup instead of changing the font. That's a better way to do it under GTK+ anyway, the only pity is that it can't be used on all platforms.

Thanks for your analysis and hunting down the occurrences!

PR #437, opened 2017-09-25
Adding ja comma to isEquivalent

chason:
In Japanese, the regular comma used includes whitespace at the end of
the character since it would normally be a full width character. In
English, a comma has whitespace at the end, but its a separate
character. This change makes it so that the IsEquivalent check
recognizes ", " and "、" as functionally the same set of characters.
The reason this is different than the full width period, is that a
translation that ends with a comma in English needs the space afterwards
which is not true for periods.

in reference to #issue 437

### DIFF
--- a/src/qa_checks.cpp
+++ b/src/qa_checks.cpp
@@ -241,6 +241,8 @@ private:
                     return trans == L'（';
                 case ')':
                     return trans == L'）';
+                case ', ':
+                    return trans == L'、';
                 default:
                     break;
             }


GitCop:

 * Commit: 897d8cd9ef8220d5e132b9f0a686d7263972c574

---

GitCop:

 * Commit: 2285e35365144717ce26ecece5c0c41b4d12b8f5

---

timofonic:
@chason ご協力 いただき ありがとう ございます

@vslavik Any news about this? No reply since months, only from a bot...

vslavik:
@timofonic Please refrain from drive-by bugging. You're not the submitter, it's not your place to complain.

@chason I didn't comment because the PR is fine and I will merge it into the next release. As you can probably imagine, this has been a busy time with the new macOS version released on the day you submitted. Thanks!

vslavik:
OK, sorry, I wasn't paying enough attention apparently and missed the space after `,`. I'm rather amazed that this code, where single characters are checked, even compiles, but it certainly does not *work* as you mean it to.

I committed a more limited (but syntactically correct) version in 7c4ceb639966b77deea901929076680fa6e40e4a. If you think fixing the `, ` case properly is worth doing, please submit another PR for it. But honestly, IMHO it's not worth doing because this situation, when a string ends with `, `, is going to be *extremely* rare — not least because it's a totally wrong approach to translation to begin with.

PR #464, opened 2018-03-22
Update appstream data to components version 0.11

gusnan:
This updates the appdata file to components version 0.11 - the one problem still in this version is the id field - running

appstreamcli validate poedit.appdata.xml

gives:

W - poedit.appdata.xml:poedit:4
    The component ID is not a reverse domain-name. Please update the ID and that of 
    the accompanying .desktop file to follow the latest version of the Desktop-Entry 
    and AppStream specifications and avoid future issues.

### DIFF
--- a/poedit.appdata.xml
+++ b/poedit.appdata.xml
@@ -1,35 +1,32 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<application>
-  <id type="desktop">poedit.desktop</id>
-  <metadata_license>CC0-1.0</metadata_license>
-  <project_license>MIT</project_license>
-  <name>Poedit</name>
-  <summary>GUI editor for GNU gettext .po files</summary>
-  <summary xml:lang="de">Grafischer Editor für GNU Gettext-Dateien</summary>
-  <description>
-    <p>
-      This program is a GUI frontend to GNU Gettext utilities and a catalogs 
-      editor/source code parser. It helps with translating applications into 
-      other languages.
-    </p>
-    <p xml:lang="de">
-      Dieses Programm stellt eine grafische Benutzeroberfläche für die
+<?xml version="1.0" encoding="utf-8"?>
+<components version="0.11">
+  <component type="desktop-application">
+    <id>poedit.desktop</id>
+    <name>Poedit</name>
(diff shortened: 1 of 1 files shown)


vslavik:
> the one problem still in this version is the id field 

So why didn't you fix that too? What's the point of doing the rest, but not this?

Also, what's the implication for older systems — i.e. since when is this format widely supported by distros?

gusnan:
> So why didn't you fix that too? What's the point of doing the rest, but not this?

I believe it is something for the upstream developer to decide, and not anyone else. I haven't found a reverse domain-name anywhere, so I thought it might be better to ask to get a correct one.

> Also, what's the implication for older systems — i.e. since when is this format widely supported by distros?

Hmm, that's a good point, maybe I am too much in a hurry with this - I am not really sure.

vslavik:
> I believe it is something for the upstream developer to decide, and not anyone else.

You're making a contribution to the upstream, it seems reasonable to me that just as you proposed this reasonable change, you're well within scope to propose other ones ;)

> I haven't found a reverse domain-name anywhere

Poedit's website is at https://poedit.net (which is linked in the patch too) — so that's the canonical domain. On macOS, Poedit uses `net.poedit.Poedit`, so if that's an appropriate format for AppData too, let's use that.

> Hmm, that's a good point, maybe I am too much in a hurry with this - I am not really sure.

Presumably you have some motivation for this PR, driven by some knowledge that I lack, and would be a better judge of this than me with zero knowledge...

vslavik:
Let me clarify my concern: it's almost certainly safe for distro-provided versions of Poedit, *if* I merge this into 2.1, not 2.0.x: it would then be bumped only with the rest of the distribution. I'd *also* like to start providing Snap and Flatpak packages, so would like to remain compatible with reasonable targets for those (which I admit I don't know offhand, but I think it might be something like $current-2 for Ubuntu/Fedora).

gusnan:
Sounds good to me - I am coming from the viewpoint of Debian unstable - where lintian (a Debian package checker) gives warnings and errors on poedit regarding its appstream data:

https://lintian.debian.org/tags/appstream-metadata-legacy-format.html
https://lintian.debian.org/tags/appstream-metadata-in-legacy-location.html

Issue #150, opened 2015-02-24
More Export Option


renemaas:
- CSV
- Excel
- XML
- JSON

vslavik:
Excel import/export is being considered. No plans for other, especially export-only or ill-defined (such as “XML” or “JSON”) things.

JToward:
Is there any import method that has been already implemented?

PR #468, opened 2018-04-23
Snap packaging

Lin-Buo-Ren:
Currently not buildable due to missing Boost dependency etc. , the packages from the PPA might help.


GitCop:

 * Commit: ece1a88eeeaf0092bea919013bacfb6de5dab7d5

---

GitCop:

 * Commit: 06fc20aaa9e4b132a93a2c60b00abc5e244ef051

---

vslavik:
> Currently not buildable due to missing Boost dependency etc. , the packages from the PPA might help.

All dependencies are in the `deps` directory and it is preferable to build using them instead of some out-of-date PPAs. In particular, wxWidgets should absolutely be used from there instead of from Ubuntu's old packages.

I'm also a bit confused about why you submit something you yourself say is non-functional as a PR? The traditional way is to work on your git branch and submit a PR only when it is ready to be merged...

Lin-Buo-Ren:
> All dependencies are in the deps directory and it is preferable to build using them instead of some
> out-of-date PPAs.

I've successfully build the dependencies using the deps directory(though still barely launch-able), will push it tonight

> I'm also a bit confused about why you submit something you yourself say is non-functional as a PR?
> The traditional way is to work on your git branch and submit a PR only when it is ready to be merged...

I hope to be noticed to avoid rebuilding wheels as I recently saw you mentioning it in another issue, anyway I'm fine with the tradition.

vslavik:
> will push it tonight

I'm even more confused: then why did you close this PR? You can use `git push` or `git push --force` to update this one...

Lin-Buo-Ren:
@vslavik Just be a bit too coward, no worries

Lin-Buo-Ren:
I've prepared and pushed my current progress.  

What's working:

* Launch (with a few by-passable errors)
* Load file
* Save file

Not working:

* input methods
* translation memory related errors
* probably lots of stuff here and there

Many problems probably related to incomplete desktop-gtk3 part integration, still looking at it.

Lin-Buo-Ren:
I noticed that there's a template repo to track the progress:
https://github.com/snapcrafters/fork-and-rename-me

I wonder if it's sane to follow the snapcrafter's protocol instead?

vslavik:
> to follow the snapcrafter's protocol instead?

Meaning what, exactly?

Lin-Buo-Ren:
> Meaning what, exactly?

I mean follow the instructions from the snapcrafters template, I am like, jumped to the 20th task on the list or something.

vslavik:
Just repeating what you said using different words doesn't explain anything I'm afraid. I am opposed to having it as separate repo if that's what you're asking. If it is not, I have no idea what you *are* asking :(

vslavik:
> switch dep. of wx to gtk2 

That's not a good idea. Why?

> Crowdin integration(The auth webpage is triggered, but the result isn't sent back to the application)

Probably some extra permission is required to pass the `poedit://` protocol through? For normal deployment, this is defined in the `poedit-uri.desktop` file.

> snappy: Improve boostlib build recipe  …
> snappy: Improve wxWidgets build recipe 

Is there no way to build from the *git submodule* associated with this particular checkout? 

Also, it makes sense to use static packages here...

Lin-Buo-Ren:
Thanks for reviewing.

>> switch dep. of wx to gtk2
>> 
> That's not a good idea. Why?
> 

The program crashes on launch at my end when using GTK3 for an unknown reason and *at the time* I thought that the checked out wxWidgets is GTK2 only, will drop it once the crash issue is sorted out.

> For normal deployment, this is defined in the poedit-uri.desktop file.

I did notice the other desktop entry file, will try to integrate it when I find out how to do it properly.

> Is there no way to build from the git submodule associated with this particular checkout?

It's definitely possible, looking into it.

> Also, it makes sense to use static packages here...

I'm not completely sure how to achieve this, but will check it out.

Lin-Buo-Ren:
Apologies for the ignorance.

> I think (based on docs) that this can be fixed by adding desktop-legacy to the snap's plugs list.

I did notice the doc but didn't trust it due to the lacking mention in `snapcraft define desktop-gtk3`, I've asked on the forum and [got a confirmative answer](https://forum.snapcraft.io/t/cant-use-input-method-in-snap-apps/4712/22?u=lin-buo-ren) today, will add it as soon as possible.

> `network` entitlement is also needed for Crowdin.

Will do.

>> translation memory related errors
>> 
> Can you be more specific?
> 

I'll note it down once I encounter it again.

GitCop:

 * Commit: d5b70feab43b3168a2acb1fe0444ffb4bf646c46

 * Commit: 8ee4eee06fdec12f202292f8897ecb86a7223921

 * Commit: 29c4bad952b721eff75d0070c8619d9a0cf4edbc

 * Commit: 61bdefea9303dd4c2c7c3ce01353edbae54c09d1

 * Commit: bd7a77f840f51d2255e29bc4874ac36eaa466891

---

snappy/.gitignore
```diff
@@ -0,0 +1,12 @@
+# Version Tracking Ignore Rules for Snappy
+# https://git-scm.com/docs/gitignore
+## Ignore built snap packages
+*.snap
+
+## Ignore stages autogeneraged files
+/*/parts/
```
vslavik: `.gitignore` in a subdirectory should use relative, not absolute (starting with `/`) paths to ignorable files...

snappy/poedit-git/snap/snapcraft.yaml
```diff
@@ -0,0 +1,176 @@
...
+      - libxt-dev
+      - libgstreamer1.0-dev
+      - libgstreamer-plugins-base1.0-dev
+      - libgtk-3-dev
+      - libwebkit2gtk-4.0-dev
+      - libnotify-dev
+    override-build: |
+      ./configure
+      make -j$(nproc)
+      sudo make install
```
vslavik: There's no reason build should use `sudo`, I'm pretty sure.

vslavik:
> ./src/common/iconbndl.cpp(359): assert ""icon.IsOk()"" failed in AddIcon(): invalid icon

That would indicate either misinstallation (some icon files omitted) or Poedit being unaware of its install location. That can be fixed by setting `POEDIT_PREFIX`, but it's probably better to figure out where in the snap things went wrong.

> wxWidgets stacktrace not viewable in the aforementioned error dialog, with the following message 

Including the binary in the snap would be one fix, but I think this is OK to ignore. I'll just disable this feature, it's not terribly useful anyway.

> oedit crashes during launching : GLib-GObject-WARNING **: specified class size for type 'IBusIMContext' is smaller than the parent type's 'GtkIMContext' class size

That's pretty clear sign of miscompilation: different, binary-incompatible, version of GTK+ used to compile wxGTK and Poedit, and/or a runtime incompatible binary (which shouldn't normally be possible to link, but well).

I'll try to have a look at those issues myself if you don't beat me to it.

I also don't understand why you have two snaps there — instead of having one for the last release and one for git `master`, there should be only one that works with the *current checked out tree*, i.e. `.`, i.e. doesn't download anything from anywhere and builds from the sources around it.

snappy/.gitignore
Lin-Buo-Ren: Refer to gitignore(5):

Patterns read from a .gitignore file in the same directory as the path, or in any parent directory, with patterns in the higher level files (up to the toplevel of the work tree) being overridden by those in lower level files down to the directory containing the file. **These patterns match relative to the location of the .gitignore file.** A project normally includes such .gitignore files in its repository, containing patterns for files generated as part of the project build.

I also have confirmed that the ignore rules are effective.

snappy/poedit-git/snap/snapcraft.yaml
Lin-Buo-Ren: This will be fixed in the next push.

Lin-Buo-Ren:
> I'll try to have a look at those issues myself if you don't beat me to it.

Thanks.

> I also don't understand why you have two snaps there — instead of having one for the last release and one for git master, there should be only one that works with the current checked out tree, i.e. ., i.e. doesn't download anything from anywhere and builds from the sources around it.

I'll attempt to return to use the in-tree dependencies and drop the additional package as soon as possible.  The pre-release builds should probably go into separate channels (edge, candidate) instead.

Lin-Buo-Ren:
I have `--enable-xrc` the `wxwidgets` parts but the configure script of Poedit didn't detect it, any ideas?

```
checking for wx-config... /home/ubuntu/poedit/snappy/parts/poedit/install/../../wxwidgets/install/bin/wx-config
checking for wxWidgets version >= 3.0.3 (--unicode)... yes (version 3.1.1)
checking for wxWidgets static library... yes
checking if wxWidgets includes XRC... no
configure: error: XRC is required to build poedit!
Failed to run './configure --prefix=/home/ubuntu/poedit/snappy/parts/poedit/install --with-wx-prefix=/home/ubuntu/poedit/snappy/parts/poedit/install/../../wxwidgets/install --with-boost=/home/ubuntu/poedit/snappy/parts/poedit/install/../../boostlib/install' for 'poedit': Exited with code 1.                         
Verify that the part is using the correct parameters and try again.
```

vslavik:
> I have --enable-xrc the wxwidgets parts but the configure script of Poedit didn't detect it, any ideas?

Sounds like you *don't*… As with any `configure` failure, `config.log` will tell you more.

This could be related to the asserts/crashes above and my suspicion about a mismatch between wx versions. That would explain why you have XRC enabled (one copy of wx), but not detected (another copy of wx).

Lin-Buo-Ren:
>  As with any configure failure, config.log will tell you more.

Here's the config.log in the wxwidgets part's build dir: http://paste.ubuntu.com/p/QVrhjq4b33/

> This could be related to the asserts/crashes above and my suspicion about a mismatch between wx versions. That would explain why you have XRC enabled (one copy of wx), but not detected (another copy of wx where it isn't enabled).

I've dropped WxWidgets system libraries from my build container quite a while now, and the Poedit configure log does reflect my build configuration in the wxwidgets part (like switching from shared lib to static lib).  I'll check if there's anything wrong in my container.

vslavik:
> Here's the config.log 

The idea was that *you* should read it to figure out what you're doing wrong — and the relevant one is of course for the configure run that is *failing*, i.e. Poedit's, which tells you *why* it failed to detect XRC libraries.

Lin-Buo-Ren:
Thanks for the explanation, here's the proper config.log from the poedit: http://paste.ubuntu.com/p/fgp3zKkd3H/

```
configure:8327: checking if wxWidgets includes XRC
configure:8346: g++ -std=gnu++14 -c  -I/home/ubuntu/poedit/snappy/stage/include -pthread -I/home/ubuntu/poedit/snappy/parts/poedit/install/../../boostlib/install/include -pthread -I/usr/include/libsecret-1 -I/usr/include/glib-2.0 -I/usr/lib/x86_64-linux-gnu/glib-2.0/include -I/home/ubuntu/poedit/snappy/parts/poedit/install/../../wxwidgets/install/lib/wx/include/base-unicode-3.1 -I/home/ubuntu/poedit/snappy/parts/poedit/install/../../wxwidgets/install/include/wx-3.1 -D_FILE_OFFSET_BITS=64 -DWXUSINGDLL -pthread  -I/home/ubuntu/poedit/snappy/stage/include conftest.cpp >&5
conftest.cpp:36:22: error: #error "XRC not compiled in"
                     #error "XRC not compiled in"
                      ^
configure:8346: $? = 1
```

Not sure why the XRC support is not compiled in... maybe some other wxWidgets configuration option conflicts with `--enable-xrc` ?

and here's the dpkg package database info in the container: http://paste.ubuntu.com/p/H7k4hMQHMD/ , just for your info.

vslavik:
> Not sure why the XRC support is not compiled in... maybe some other wxWidgets configuration option conflicts with --enable-xrc ?

No. 

As I keep telling you, and you keep ignoring, you have a mess in your tree with multiple conflicting copies of wx. I can't remotely debug things you don't disclose, you need to dig in and investigate. `/home/ubuntu/poedit/snappy/stage/include` is the obvious first place to look in. 

That you *do* have a mess is immediately obvious from the two logs above: the wx one creates `setup.h` (the configs storing headers) use `lib/wx/include/base-unicode-static-3.1/wx/setup.h`, your poedit checks look in `lib/wx/include/base-unicode-3.1`. You have *at least* two differing versions of wx in there and not using the one you think you're using.

Wipe out everything in build directories. Start a fresh build.

vslavik:
Returning to this: I initially assumed that snap build process would run from a clean slate, as e.g. RPM builds do, and must be interfering with something system-wide. But if the temporary files *are* kept, then here's what could plausibly have happened: you initially build shared version of wx, then after my feedback changed to static, which caused your build directory to contain two copies; the older shared one apparently takes priority.

td;dr: deleting any build directories there might be actually *really* should help, if the above theory is correct.

Lin-Buo-Ren:
As always, apologies for the nuisance.  At the time the poedit configure log is collected I was switching between the static and non-static build of the wxwidgets part without rebuilding the poedit part so it might be the case, here's the history:

```
  226  snapcraft clean wxwidgets -s build
  227  snapcraft build wxwidgets
  228  snapcraft stage wxwidgets
  229  snapcraft build poedit
  230  ls
  231  ./parts/wxwidgets/install/bin/wx-config 
  232  snapcraft clean wxwidgets -s build
  233  snapcraft build wxwidgets
  234  tree parts/wxwidgets/install/
  235  man tree
  236  tree parts/wxwidgets/install -C | less --raw-control-chars
  237  wx-config
  238  less parts/wxwidgets/build/config.log 
  239  pastebinit parts/wxwidgets/build/config.log 
  240  dpkg --status
  241  dpkg --list
  242  dpkg --list | grep wx
  243  dpkg --list | pastebinit
  244  pastebinit parts/poedit/build/config.log 
  245  dpkg --list
  246  dpkg --list | pastebinit
```

I rather noticed the fact that the stage directory *shouldn't really have the wx and boost headers at all in the first place* as we are static-linking these libraries from the parts installdir without making them into the the stage and even the prime phase.  The next push will drop these unexpected files out of the stagedir.

vslavik:
> As always, apologies for the nuisance

Not at all!

Lin-Buo-Ren:
### Current Status
* Poedit no longer crashes at launch due to input method error
* Input method (iBus) works now
* Stacktrace printing in the wxWidgets error dialog works now
* Application icon in the menu is properly displayed

### Known Issues
* Icon loading errors (probably due to GTK theme?)
* Crowdin integration not working
* Errors in system log: https://pastebin.com/0fApnFJi
* Errors in stdout/stderr: https://pastebin.com/4x5neSJa

vslavik:
Question: are you getting prompts about missing spellchecking dictionaries in Poedit snap? Per https://forum.snapcraft.io/t/desktop-allow-access-to-host-hunspell-dictionaries/2598 this will require extra work :(

vslavik:
OK, I can't keep up with this. Please just tell me when you think it's ready to be reviewed.

Lin-Buo-Ren:
I've attempted to make the URL handler work but to no avail, here are some details:

1. The desktop entry that is actually in effective is installed at  
`/var/lib/snapd/desktop/applications/poedit_poedit-url-wrapper.desktop`  
whose `Exec` key has been replaced to:  
`env BAMF_DESKTOP_FILE_HINT=/var/lib/snapd/desktop/applications/poedit_poedit-uri-handler.desktop /snap/bin/poedit.poedit-uri-handler %U` (instead of `%u` in the original desktop entry)
1. /snap/bin/poedit.poedit-uri-handler is a symbolic link pointed to /usr/bin/snap, which probably indirectly runs `$SNAP/command-poedit-uri-handler.wrapper`
1. Which is a shellscript that in the end runs:  
`exec "desktop-launch" "poedit" "--handle-poedit-uri" "$@"`

The iso-codes related error messages:

```
** (poedit:26466): WARNING **: iso_639.xml: Failed to open file '/usr/share/xml/iso-codes/iso_639.xml': open() failed: No such file or directory

** (poedit:26466): WARNING **: iso_3166.xml: Failed to open file '/usr/share/xml/iso-codes/iso_3166.xml': open() failed: No such file or directory
```

roots from the gtkspell-codetable.c code from gtkspell3, which seems to be not honoring `XDG_DATA_DIRS` and thus unable to open the data files provided by the iso-codes packages staged in `stage-packages`.

This might be the reason why Poedit is unable to find the spellcheck dictionaries even when they are packed into the snap and the cause of the following error message from gtkspell.c:

```
** (poedit:26466): CRITICAL **: check_range: assertion 'spell->priv->speller != NULL' failed
```

vslavik:
> I've attempted to make the URL handler work but to no avail, here are some details:

You describe how Snap implements the wrapper, all of which seems reasonable to me, without saying what the problem *is*, so it's hard to advice… This seems suspicious, though:

> command: desktop-launch poedit --handle-poedit-uri

This should be just the normal command I think. IOW, I suspect the *actual* command being run may be more like `exec "desktop-launch" "poedit" "--handle-poedit-uri" "--handle-poedit-uri" "$@"`

Lin-Buo-Ren:
I'm currently hunting down the cause of the not-available spellchecking functionality, this seems to be due to the non-relocatable status in the Enchant library (i.e. only search required resources under /usr instead of /snap/poedit/current/share/...

As the underlying library needs to be patched static linking is no longer preferable as all the intermediary libraries (Gtkspell) need to be built from source instead of only the problematic ones(Enchant).

The relocatability of each parts can be verified by running `snap run --strace='-o poedit.stracerace=execve,access,open,close,fstat,stat,write' poedit` check the strace logs.

vslavik:
> I'm currently hunting down the cause of the not-available spellchecking functionality, this seems to be due to the non-relocatable status in the Enchant library

*Current* versions of Enchant — which Ubuntu doesn't package — are relocatable. The old one supports setting `DICPATH`, but the catch is that there [must be only a single directory in it](https://github.com/AbiWord/enchant/commit/93faf7fe371b0a403d5cd4e745f90758fd546f6d). So that's probably why it doesn't work, because you set it to multiple paths:

```
DICPATH="$SNAP/usr/share/hunspell:$SNAP/usr/share/myspell/dicts"
```

Changing that to only one of them (I don't know which one is the correct one) should help, hopefully.

Lin-Buo-Ren:
> Current versions of Enchant — which Ubuntu doesn't package — are relocatable. The old one supports setting DICPATH, but the catch is that there must be only a single directory in it. So that's probably why it doesn't work, because you set it to multiple paths:
> 
>     DICPATH="$SNAP/usr/share/hunspell:$SNAP/usr/share/myspell/dicts"
>
Thanks for the info, unfortunately, the issue can still be reproduced after fixing `PICPATH`.  Also according to the [strace](https://github.com/vslavik/poedit/files/2071874/poedit.strace.log) the Enchant library doesn't seem to touch the spellchecking dictionaries and even their directories *at all*([control case trace](https://github.com/vslavik/poedit/files/2071873/poedit.control-case.strace.log)).

I found the following passage in Enchant's README:

>    Enchant is also meant to be used in a cross-platform (XP) environment. Part
of this means that Enchant wants to limit its number of external dependencies
to 0, or as close is as humanly possible. Also, any enchant consumer (i.e. a
Word Processor) should not need to know about what backend providers Enchant
knows about. In fact, Enchant shouldn't even need to know this information
itself. To accomplish this, all of Enchant's providers are DLLs.

Maybe Enchant will only work when it is dynamically linked?

vslavik:
> Maybe Enchant will only work when it is dynamically linked?

I don't even see any evidence of Enchant being built statically in this PR. But if there's some such form of it that requires statically linking in providers, then some special care need to be taken to link them it. 

When I said thinks should be linked statically, I was of course referring to Poedit itself and *it's* atypical dependencies. Not to base system staff that comes form packages.

vslavik:
> I've resolved the missing spell checking dictionaries issue, the last obvious issue in the snappy build is the crowdin integration support.

I.e. the problem with `poedit://` links, which can worst case be solved by some code changes — I'll look into that one myself.

Thanks!

Lin-Buo-Ren:
I've done all that I can do to the packaging and would like to request a review.  This branch has been rebased to commit 03661c1 .

Lin-Buo-Ren:
It seems that the guys behind the snap store pre-registered all well-known package names from being taken away too easily, you might want to [filing a dispute to take over the name](https://dashboard.snapcraft.io/register-snap/?series=16&name=poedit) if you haven't.

vslavik:
> It seems that the guys behind the snap store pre-registered all well-known package names from being taken away too easily, you might want to filing a dispute to take over the name if you haven't.

Wow, nothing is easy with Snapcraft, is it? Thanks for bringing this to my attention — done now.

And sorry for the delay in reviewing this,I got sidetracked into reading up on the Snapcraft limitations. In retrospect, I should have just merged this as-is and worry about potential improvements later.

vslavik:
How do you actually *build* this? I spent several hours trying to understand snapcraft and still can't get anywhere near to compiling the snap:

1. If I use `snapcraft build` in Ubuntu, I get asked for sudo access — apparently, it wants to install system-wide stuff, which, uh, no.

2. I can't build with docker (per docs) either:

```
$ docker run -it -v "$PWD:$PWD" -w "$PWD" snapcore/snapcraft snapcraft

…

'ccache' has prerequisites that need to be staged: gcc-6
Staging gcc-6 
Failed to copy '<path_to_source_tree>/parts/gcc-6/install/bin/cpp': no such file or directory.
Check the path and try again.
```

Lin-Buo-Ren:
> How do you actually build this?
[I use a specifically crafted LXD container to do so](https://forum.snapcraft.io/t/how-to-create-a-lxd-container-for-snap-development/4658), `snapcraft` will also do the similar thing with the `cleanbuild` command option.

> apparently, it wants to install system-wide stuff, which, uh, no.

Yep that it's designed behavior, that's why building in an LXD/Docker container is recommended.

> I can't build with docker (per docs) either:
> Failed to copy '<path_to_source_tree>/parts/gcc-6/install/bin/cpp': no such file or directory.

This shouldn't happen, I'll try to reproduce it to see what's the cause.

Lin-Buo-Ren:
> I can't build with docker (per docs) either:
> Failed to copy '<path_to_source_tree>/parts/gcc-6/install/bin/cpp': no such file or directory.

`$SNAPCRAFT_PART_INSTALL/bin/cpp` is a symbolic link points to `/usr/bin/cpp-6` to make snapcraft force using the version from GCC6 and is created during the build time:

https://github.com/Lin-Buo-Ren/poedit/blob/wip-snappy-packaging/snap/snapcraft.yaml#L437-L465

Clean the gcc-6 part and rebuild should fix the error.

Lin-Buo-Ren:
I'm looking into dropping the asciidoc dependency as it pulls a lot of texlive packages just to seemingly build a manpage that is not accessible by the user.

Lin-Buo-Ren:
The high object count in this repo (72282 at depth 80) is quite a concern to me as it makes every clone lengthy, it might be helpful if more bundled dependencies are moved to the independent submodules.

vslavik:
> Clean the gcc-6 part by running snapcraft clean --step=build gcc-6 and rerun snapcraft should fix the error.

That's a nice theory, but no. Clean builds, repeatedly, including the very first one, do this. I don't think `cpp` is actually needed in a new version, so I'm currently trying without that.

> I'm looking into dropping the asciidoc dependency as it pulls a lot of texlive packages just to seemingly build a 

That's a fair point. My younger self didn't want generated files in the repository, but I'm much less orthodox about such things these days and the man page wasn't updated since 2008. I'll loose the dependency.

> manpage that is not accessible by the user.

I continue being surprised by how half-baked snap is :( 

> The high object count in this repo (72282 at depth 80) is quite a concern to me as it makes every clone lengthy,

With sane tools (which snapcraft, I'm becoming more and more convinced, may not be quite yet…) you clone *once* — not a big deal then. Removing content now won't change it either. But yes, I am well aware of the existence of submodules, as you might deduce from their use; from that, you might also deduce there might be *reasons* for having the things aren't in submodules as copies in a tree — and indeed, there are or were.

vslavik:
> so I'm currently trying without that.

…and that was naive:

```
Preparing to build ccache 
Building ccache 
Staging ccache 
Failed to copy '/Users/vaclav/Development/poedit2/parts/ccache/install/bin/gcc': no such file or directory.
Check the path and try again.
```

I don't know WTH its problem is. I'm running in Docker, on macOS (i.e. UNIX), symlinks are a thing that works, _and_ I was getting this in Ubuntu VM as well.

Next up: trying to make a copy instead, in case across-volumes symlink just don't always work… That seems to do the trick.

Lin-Buo-Ren:
I'll probably drop the entire ccache thing in the next push, the whole point of using ccache is to speed up development of the packaging, as the implementation is close to complete it is no longer help but complicate things.

vslavik:
Do you by any chance understand why some child window (file open, about window) don't have any shadows in the snap, while some (preferences) and the main window do? 

parallels picture

Lin-Buo-Ren:
I have no idea about that, here's a comparison at my end:

default

Lin-Buo-Ren:
Here's a screenshot produced in GNOME desktop environment:



Is the about and file open dialog native Gtk dialogs?

Lin-Buo-Ren:
I've fowarded the issue to the seemingly responsible project:  
[[Gtk3] Window decoration defects with native Gtk dialogs · Issue #150 · ubuntu/snapcraft-desktop-helpers](https://github.com/ubuntu/snapcraft-desktop-helpers/issues/150)

snap/snapcraft.yaml
```diff
@@ -82,12 +82,11 @@ parts:
       - patches
       - wxwidgets
 
-    source: git://github.com/vslavik/poedit.git
+    source: .git
```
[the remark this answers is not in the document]
Lin-Buo-Ren: The `source-type: local` (when specifying `source` as `.`) copies all non-snapcraft-specific(/snap, /parts, /stage, and /prime) files and folders to the part's `src` folder (`/parts/_part_name_/src`), including the entire `/.git` folder and ignored files of the VCS.

When using `source-type: git` `snapcraft` only clones and check out the repository specified by `source` keyword, which is considered cleaner than `source: .`.

vslavik:
> Is the about and file open dialog native Gtk dialogs? 

Yes, exactly.

> Smells like bug in the snapcraft-desktop-helpers's gtk3 support.

:(

vslavik:
I have an update on the Crowdin links: Snapcraft pretty much ignores the command line and does not expand `%u` in the correct place. I'm writing a fix for it, but I think it's only part of the problem — even if the code gets through, it doesn't remember the auth token. It looks like libsecret is failing (and because Poedit's error reporting sucks, there's no visible information about it). This may be related to this error I get on every launch (i.e. before accessing libsecret and yes, I connected the snap per the instructions):

```
Failed to connect to session manager: None of the authentication protocols specified are supported
```

Can't understand it yet, it seems everybody is getting this error and blaming pretty much anything on it, with typical shotgun debugging advice on how to address. Do you get the error on launch too?

snap/snapcraft.yaml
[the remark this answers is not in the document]
Lin-Buo-Ren: That makes sense, thanks for the explanation!  Note there are [`source-commit`](https://docs.snapcraft.io/build-snaps/syntax#source-commit) and [`source-checksum`](https://docs.snapcraft.io/build-snaps/syntax#source-checksum) keywords to have further control over the checked out part sources.

Lin-Buo-Ren:
> Do you get the error on launch too?

Pretty much no, I might have seen this in other snap but not this one.

Lin-Buo-Ren:
I just gave you write access to my forking repo, you might want to force push your progress here (if you're into it, of course).

vslavik:
> Failed to connect to session manager: None of the authentication protocols specified are supported

Reading through wxGTK sources, this is due to a failed `SmcOpenConnection()` call, which is only used to detect whether we're running under GNOME/KDE, so it seems pretty unimportant.

vslavik:
> I just gave you write access to my forking repo, you might want to force push your progress here (if you're into it, of course).

By default PR branches can be pushed by repo owner already, but I instead pushed it to a branch in this repo: https://github.com/vslavik/poedit/tree/linux-snap 

I also published it to the store (http://snapcraft.io/poedit), meaning you can safely remove `poedit-wip-pr-468` from it.

If you don't see anything wrong with the snap, I'll merge it. Right now I'm aware of two small issues: 1. only for amd64 (I'll address that) and 2. in Software Center, the package is showing as "poedit" even though the metadata name is "Poedit 2" and even though your snap, as well as anything else in the store, is using the correct name... 🤷‍♂️

Lin-Buo-Ren:
> meaning you can safely remove poedit-wip-pr-468 from it.

I'll unpublish that snap soonly after the last obsoletion notice update.

> If you don't see anything wrong with the snap, I'll merge it.

I've installed the snap and tested a bit, no (new) issues noticed so far.  Go ahead if you wish.

> 2. in Software Center, the package is showing as "poedit" even though the metadata name is "Poedit 2" and even though your snap, as well as anything else in the store, is using the correct name... 

Mine as well (KDE Discover 5.11.0, KDE neon):

screenshot_20180829_134134

I doubt the current snap store backend support fetching the store metadata...

Also note: [Snap title and icon customizations seemed to be overwritten by newly published snap · Issue #746 · canonical-websites/snapcraft.io](https://github.com/canonical-websites/snapcraft.io/issues/746)

vslavik:
> I'll unpublish that snap shortly after the last obsoletion notice update.

Thanks! 

> I doubt the current snap store backend support fetching the store metadata…

It has to get them from somewhere and it's not the full huge snap package… I think this is just a caching issue, things went back to normal after I logged off and in again. 

> Also note: Snap title and icon customizations seemed to be overwritten by newly published snap

FWIW, I'm not seeing that.

---

I have now merged the snap support in 7b902ba6d7a8a87c83295cb6c3b14f10b6896e65 — the content of this PR being squashed into 2d80574757e0f4680864146639c40a7c177b2944. Thanks a lot for your help with this!

vslavik:
Yeah, I should… I'm giving it a bit of time for beta testing by more users, to see if there are any issues, and will update information about it a bit later (~a week or so I think, depending on how it goes).

PR #470, opened 2018-04-25
Implement EditorConfig(http://editorconfig.org)

Lin-Buo-Ren:
"EditorConfig helps developers define and maintain consistent coding
 styles between different editors and IDEs."

This patch implments a simple EditorConfig to force tab indentation
on /.gitmodules in complying editors.  It also copied some existing
rules in .gitattributes.

Refer-to: EditorConfig <http://editorconfig.org>
Signed-off-by: 林博仁(Buo-ren Lin) <Buo.Ren.Lin@gmail.com>

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,14 @@
+# EditorConfig is awesome: http://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+[*.sh]
+end_of_line = lf
+
+[*.{sln,vcproj,vcxproj*}]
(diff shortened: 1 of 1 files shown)


vslavik:
Having EditorConfig rules describing formatting of Poedit's source code would be useful. I don't think this is and will not merge this sort of extremely limited rules — it would only confuse EditorConfig-aware editors into assuming there's support for it, while for all practical purpose (i.e. editing actual source code) there isn't.

PR #477, opened 2018-05-15
Patch gitignore

Lin-Buo-Ren:
This pull request adds a few tracking ignore rules to the gitignore(5)


vslavik:
FYI, I didn't apply this yet because I have mixed feelings about it: the recommended way to build using Autoconf is out-of-tree, i.e. have a dedicated directory with only build files.

PR #492, opened 2018-06-08
Travis: Incorporate Compiler cache(ccache)

Lin-Buo-Ren:
This patch enables ccache caching for a faster rebuild.

Without cache:

The build time increases about 90 seconds more than when ccache is not enabled(approx. 11m50s).

With existing cache([log](https://travis-ci.com/Lin-Buo-Ren/poedit/jobs/128351224)):

the build time dropped to about 7 minutes.

The build results in a 35.38MiB of cache data, it *might* be sane to further restrict ccache size to 50MiB to avoid constantly increasing cache size in subsequent builds(by default the cache size limit is 5.0GiB)

Signed-off-by: 林博仁(Buo-ren Lin) <Buo.Ren.Lin@gmail.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,11 +34,11 @@ addons:
             - libsecret-1-dev
 
 env:
-    - REAL_CC=gcc-6 REAL_CXX=g++-6
+    - REAL_CC='ccache gcc-6' REAL_CXX='ccache g++-6'
 
 before_install:
     # work around Travis being difficult, see https://github.com/travis-ci/travis-ci/issues/6633
-    - export CC=$REAL_CC CXX=$REAL_CXX
+    - export CC="$REAL_CC" CXX="$REAL_CXX"
     # git submodules needed for Linux:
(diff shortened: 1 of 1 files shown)


vslavik:
Thanks!

(BTW, please note how I edited the commit message, it's a common pattern: redundancy or explanations shouldn't be in the subject, `Signed-off-by` only makes sense if you are *not* the committer and only in projects that use that convention.)

vslavik:
I was actually surprised that Travis documents this and doesn't frown upon the practice, ccache is notorious for yielding a huge amount of cache data that will never be used again… But if they're OK with it, I'm not the one to complain.

Lin-Buo-Ren:
> BTW

Got it.

> I was actually surprised that Travis documents this and doesn't frown upon the practice, ccache is notorious for yielding a huge amount of cache data that will never be used again

They seem to be setting a sane cache size limit in the first place so it's probably not a problem.

PR #499, opened 2018-07-17
Update configure check for pugixml

c72578:
- Add check for headers and lib if
  pugixml.pc is not available

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -222,14 +222,35 @@ PKG_CHECK_MODULES([LUCENE], [liblucene++ >= 3.0.5],
         ])
 
 
-PKG_CHECK_MODULES([PUGIXML], [pugixml >= 1.9],
-        [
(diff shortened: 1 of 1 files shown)


vslavik:
Sorry, why wouldn't `pugixml.pc` be available?

vslavik:
I see.

But honestly, I'd rather not do this here — it nontrivially complicates the code (i.e. kind of eliminates the whole point of `pkg-config`). Rationale: For users compiling from sources for their own needs, it shouldn't matter than a bundled copy is used. Presumably this only matters to distros and there a better fix is to simply enable the respective option when building the pugixml package, no?

c72578:
Makes sense, not to "workaround" here at Poedit. Sooner or later a new release of pugixml will improve the situation concerning pugixml.pc

c72578:
Just for the records:
pugixml v1.10 has been released [1] and pugixml.pc is installed now by default.

[1] https://github.com/zeux/pugixml/releases

PR #512, opened 2018-09-02
Drop unnecessary PATH modification in snap build

Lin-Buo-Ren:
Snapcraft already gives `$SNAPCRAFT_STAGE/bin` priority in the build
step and doesn't require setting it manually.


vslavik:
Thanks!

PR #514, opened 2018-09-04
Update the AppData file with new information

bilelmoussaoui:
This PR updates the AppData with the latest requirements:
- OARS Rating 
- Translation tag to allow Software center to show the application as translated on the current user's language
- Issues URL
- Releases information, it lacks the old releases' data. Do you think that you can keep the releases section up to date?

### DIFF
--- a/poedit.appdata.xml
+++ b/poedit.appdata.xml
@@ -1,31 +1,76 @@
 <?xml version="1.0" encoding="utf-8"?>
-<components version="0.11">
-  <component type="desktop-application">
+<component type="desktop">
     <id>poedit.desktop</id>
     <name>Poedit</name>
     <summary>Translations editor for gettext</summary>
     <project_license>MIT</project_license>
+    <metadata_license>CC0</metadata_license>
     <description>
-      <p>
-        Translations editor for gettext (PO files). It helps with translating
-        applications or WordPress themes and plugins into other languages, as
-        well as with developing localizable applications.
-      </p>
+        <p>
+            Translations editor for gettext (PO files). It helps with translating
+            applications or WordPress themes and plugins into other languages, as
+            well as with developing localizable applications.
+        </p>
     </description>
     <screenshots>
-      <screenshot>
-        <image type="source">https://upload.wikimedia.org/wikipedia/commons/c/c2/Poedit_1.8.1_en.png</image>
-      </screenshot>
+        <screenshot>
+            <image type="source">https://upload.wikimedia.org/wikipedia/commons/c/c2/Poedit_1.8.1_en.png</image>
+        </screenshot>
     </screenshots>
     <url type="homepage">https://poedit.net</url>
+    <url type="bugtracker">https://github.com/vslavik/poedit</url>
     <url type="translate">https://crowdin.com/project/poedit</url>
+    <content_rating type="oars-1.0" />
+    <releases>
+        <release version="2.1.1" date="2018-07-26">
+            <description>
+                <p>Fixed breakage of some localizations on macOS.</p>
+            </description>
+        </release>
+
+        <release version="2.1" date="2018-07-23">
+            <description>
+                <ul>
+                    <li>Added import and export of translation memory as TMX files.</li>
+                    <li>Added ability to delete bad translations from the TM.</li>
+                    <li>TM now has limited support for plural forms (only nplurals ≤ 2).</li>
+                    <li>Improved handling plural form rules. CLDR is now used as the data source and expressions
+                        are checked for equivalence before warning about unusual forms.</li>
+                </ul>
+            </description>
+        </release>
+        <release version="2.0.9" date="2018-07-11">
+            <description>
+                <ul>
+                    <li>Improved dark theme supports (still not perfect).</li>
+                    <li>Fixed broken list rendering of RTL text on Windows.</li>
+                </ul>
+            </description>
+        </release>
+
+        <release version="2.0.8" date="2018-06-03">
+            <description>
+                <ul>
+                    <li>Added CakePHP support.</li>
+                    <li>Minor QA warnings and RTL fixes.</li>
+                    <li>Improved TM reset to work when the index is corrupted.</li>
+                </ul>
+            </description>
+        </release>
+
+    </releases>
     <keywords>
-      <keyword>translation</keyword>
-      <keyword>gettext</keyword>
-      <keyword>po</keyword>
-      <keyword>localization</keyword>
-      <keyword>i18n</keyword>
-      <keyword>wordpress</keyword>
+        <keyword>translation</keyword>
+        <keyword>gettext</keyword>
+        <keyword>po</keyword>
+        <keyword>localization</keyword>
+        <keyword>i18n</keyword>
+        <keyword>wordpress</keyword>
     </keywords>
-  </component>
-</components>
+    <provides>
+        <binary>poedit</binary>
+    </provides>
+
+    <developer_name>Václav Slavík</developer_name>
+    <translation type="gettext">poedit</translation>
+</component>


bilelmoussaoui:
We would like to update the screenshots on Linux too. Is it possible to host them either on the Website or here on GitHub?

vslavik:
> We would like to update the screenshots on Linux too. Is it possible to host them either on the Website or here on GitHub?

Sure, I'll upload them under https://poedit.net/screenshots/linux/

poedit.appdata.xml
```diff
@@ -1,31 +1,76 @@
 <?xml version="1.0" encoding="utf-8"?>
-<components version="0.11">
```
vslavik: What's the implication of removing this for older distros, if any? Won't it break parsing by some?

poedit.appdata.xml
```diff
@@ -1,31 +1,76 @@
...
-  <component type="desktop-application">
+<component type="desktop">
     <id>poedit.desktop</id>
```
vslavik: Since we're updating this for conformance with AppStream spec, it would be nice to update the ID as well and use `net.poedit.Poedit`.

poedit.appdata.xml
```diff
@@ -1,31 +1,76 @@
...
     <name>Poedit</name>
     <summary>Translations editor for gettext</summary>
     <project_license>MIT</project_license>
+    <metadata_license>CC0</metadata_license>
```
vslavik: The spec lists `CC0-1.0` but not `CC0` among valid licenses. While I realize this too is SPDX code, it might make sense to stick to the more specific versioned string.

poedit.appdata.xml
```diff
@@ -1,31 +1,76 @@
...
-        <image type="source">https://upload.wikimedia.org/wikipedia/commons/c/c2/Poedit_1.8.1_en.png</image>
-      </screenshot>
+        <screenshot>
+            <image type="source">https://upload.wikimedia.org/wikipedia/commons/c/c2/Poedit_1.8.1_en.png</image>
+        </screenshot>
     </screenshots>
     <url type="homepage">https://poedit.net</url>
+    <url type="bugtracker">https://github.com/vslavik/poedit</url>
     <url type="translate">https://crowdin.com/project/poedit</url>
+    <content_rating type="oars-1.0" />
```
vslavik: Poedit doesn't contain any content of its own, so adding this tag, especially empty, doesn't seem to make sense to me. Per [the spec](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-content_rating) and contrary to this PR's description, `<content_rating>` is optional and not required, so why include it? Am I missing something?

poedit.appdata.xml
```diff
@@ -1,31 +1,76 @@
...
+    <releases>
```
vslavik: I have mixed feelings about this. The usefulness of this is obvious, but I'd really like to avoid *manually* updating this information in two places (`NEWS` and this)...

poedit.appdata.xml
bilelmoussaoui: Older distros have a strict policy to update applications...

poedit.appdata.xml
bilelmoussaoui: Yeah that would be perfect! Will take care of that :+1:

poedit.appdata.xml
bilelmoussaoui: Will fix that!

poedit.appdata.xml
bilelmoussaoui: @vslavik It's not required for the spec but it's required for Flathub.

poedit.appdata.xml
bilelmoussaoui: @vslavik the NEWS file is read by devs. The release tags here will be shown in the software center which will allow the users to see if it's worth it to upgrade the software to the latest release. You can of course only push the latest two releases here and so on. No need to keep the whole history.

poedit.appdata.xml
vslavik: > Older distros have a strict policy to update applications…

This PR was motivated by a flatpak package update. The whole point of flatpak (for software like Poedit that is natively packaged) is deploying to older distros that don't have the latest & greatest natively packaged. Presumably such a bit older distros won't have flatpak runtime always at the latest version either and maybe the AppStream stuff is exposed to the hosting OS as well for integration.

So again: what are compatibility implications? How new must the tools be to understand `<components>`-less files?

poedit.appdata.xml
vslavik: Sounds like a Flathub bug to me, then — because such requirement is Majorly Stupid™ as explained above. And if this code, i.e. *value-less* tag, satisfies Flathub, that's doubly so :( But OK, so be it. An explanation that this is to keep Flathub happy would be nice, though.

poedit.appdata.xml
vslavik: I am aware. The point is, `NEWS` is the canonical location for this information, it's a de-facto place for users (yes) to look for it, so I *have* to update that one. And I'd rather not manually write the same information in more than 1 place. "No need to keep the whole history" solves nothing in that regard (indeed, it's *more* work to remove old entries on top of adding new ones).

poedit.appdata.xml
bilelmoussaoui: I'm not related to Flathub at all. Just a random geek trying to make software's easily installable on Linux :)
@nedrichards Do you have anything to say about this? Thanks

poedit.appdata.xml
bilelmoussaoui: You could somehow make the releases tags get updated automatically based on the NEWS file. I could create a simple script for that if you want and put it somewhere in the source tree.

poedit.appdata.xml
bilelmoussaoui: The appdata is used by the software center to show up the information related to the application the user wants to install. Flatpak/Flathub package will work without any issues. 
The issue here is whether that old appcenter will show up the correct appdata information or not. The example in the documentation is using the same thing as I proposed on this PR https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html#sect-Quickstart-DesktopApps which was written in 2013

vslavik: That really doesn't say much — that's some random example maybe initially written in 2013 and possibly edited since then (and the whole comment might have been made up, including the year — it's an example).

Here's a [recent file](https://github.com/ximion/appstream/blob/master/tests/samples/collection/xml/lvfs-gdpr.xml) in AppStream's repo still using `<components>`. The [current schema](https://github.com/ximion/appstream/blob/master/data/appstream.xsd) for the file format still allows for it.

But the schema was also last modified in 2014, which shows that not using `<components>` was valid back then. That's old enough for me.

It brings another question, though: why change this bit and remove `<components>`? What was the motivating reason?

bilelmoussaoui: Never said that it was outdated or deprecated. It's just that your application ships only one component. I can revert that part of course!

vslavik: Well, you did: "This PR updates the AppData **with the latest requirements**", so I operated under the assumption it is needed (because PR description and commit messages are truthful and one-thing-per-one-commit).

It does make sense to clean this up as you did, but it's probably more suitable as a separate commit (within a single PR).

bilelmoussaoui: Alright, will take care of that this evening with the rest of the comments! thanks for the review :+1:

vslavik:
The content rating stuff is even more complicated: AppStream tools complain (rightfully, IMHO) about empty tag:

```
$ appstreamcli validate --pedantic poedit.appdata.xml
P - poedit.appdata.xml:poedit.desktop:16
    The screenshot does not have a caption text.

W - poedit.appdata.xml:poedit.desktop:23
    Found empty 'content_rating' tag.

W - poedit.appdata.xml:poedit.desktop:3
    The component ID is not a reverse domain-name. Please update the ID and that of 
    the accompanying .desktop file to follow the latest version of the Desktop-Entry 
    and AppStream specifications and avoid future issues.

Validation failed: warnings: 2, pedantic: 1
```

So it's either accommodating Flatpak or being well-behaved AppStream file...

vslavik:
I wanted to get this into Poedit 2.2, because it is potentially breaking packaging and thus wouldn't be appropriate for 2.2.1. So I applied the fixes myself, but left out controversial parts — i.e. releases and content rating. 

I really don't know how to handle the latter :(

bilelmoussaoui:
The empty tag is the latest version of OARS, you could basically use all the tags with none as a value. https://hughsie.github.io/oars/

bilelmoussaoui:
Btw, I don't think it's the right file for the uri handler desktop file https://github.com/vslavik/poedit/blob/a3a6cba097e30dc1142f8c9863d1582530661d79/net.poedit.PoeditURI.desktop
It should be something like `net.poedit.Poedit.URIHandler` or something like that. Otherwise, it won't be exported within the flatpak package. 
For the releases tag, are you in having a script that does the whole job for you? converting the NEWS file to release tags?

vslavik:
> The empty tag is the latest version of OARS, you could basically use all the tags with none as a value.

That may well be, but your comment is going sideways and not addressing the actual issue. Which is: the spec says this is optional. Official AppData tools in a current distro (Ubuntu 18.04) flag empty tag with a warning. Flathub, you say, insists on this tag being present no matter the sense of it. It's currently impossible to cleanly satisfy `appstreamcli validate` and Flathub both.

> It should be something like net.poedit.Poedit.URIHandler or something like that.

Again, please corroborate such claims with links to specs.

vslavik:
> For appstreamcli validate, you should be using validate-relax instead :) validate enforces a lot f of things that are "optional". You can have more information about that here https://github.com/flathub/flathub/wiki/App-Requirements

Again: that is *not* the issue. `appstreamcli` *isn't* enforcing anything, it's warning about silly markup. *Flathub* is enforcing things not part of the spec — see your own link. Well, "enforcing" is probably a wrong word, because apparently they only *claim* to require such things but the current Poedit package is evidence to the contrary, isn't it.

So to reiterate:

1. AppStream validation tools frown upon empty `<content_rating>` and considers it correct to not have it.
2. Flathub frowns upon not having `<content_rating>` and considers it correct to have it empty. 

Neither has it as a hard requirement, apparently. It is impossible to satisfy both. Satisfying spec makes Flathub unhappy. Satisfying Flathub makes spec unhappy. Finally,

3. It makes no freaking sense to have `<content_rating>` in the first place, there's no *content* in there.

…which makes me want to side with the spec and not Flathub, because it's the more sensible position of the two. If it's possible to make both happy, that would be best, but I don't know how yet, and can't dedicate time to finding out right now.

Is it more clear now?

Regarding releases notes, as I said, I don't want to do it manually, scripts are fine (there's already `chversion` script for bumping the version that is a suitable place).

bilelmoussaoui:
You can, of course, keep the appdata file without a oars tag until things are enforced more everywhere.  It's a requirement yeah, but I guess it's more for games or apps that might use the user's data.

bilelmoussaoui:
Alright, will make a script and push it on a separate repository as a lot of open source software out there could use it :)

vslavik:
It seems unlikely that something working well for Poedit would be generally useful… For example, Poedit's `NEWS` file mixes multiple platforms, so such a script would need to filter out non-Linux items etc...

PR #519, opened 2018-09-12
Use global scope plugs assertion in the snap

Lin-Buo-Ren:
There's a poorly documented plug assertion format that applies to the
entire snap, by migrating to this we can avoid duplicating the plug
enumeration for each of the _app_name_ and also compatible with plugs
that has attributes (like the content interface).


vslavik:
I'm always for simplifications, thanks!

Issue #526, opened 2018-11-07
Shortcut for compile in MO


vaites:
There's no shorcut for this item, and if you disable the automatic compilation to MO can be very useful. 

I'm using Poedit 2.2 on macOS 10.12.6.

c72578:
There is the possibility to use: File - Compile to MO...
Did you mean, you would like to have a shortcut for this?

BTW, the name of the program is Poedit (not PoEdit)

vslavik:
> and if you disable the automatic compilation to MO can be very useful.

Can you please elaborate? *Is* it actually needed by you, and if, for what workflow? I'm confused by this because on one hand you disabled automatic compilation, i.e. you don't need this to be done hassle-free and frequently, but on the other, you want a shortcut, so presumably you _do_ want to use it frequently… This seems like a contradiction to me.

vslavik:
> The framework used in many projects works without problem with MO files so they are compiled to PO only for production.

This confused me before I realized you swapped "PO" and "MO". "MO" stands for "machine object", i.e. the compiled version.

Still: I do understand why you don't *need* MOs, but what I don't understand is why you feel the need to *disable them* — presumably Symfony can load MOs from the same location as well, so I don't see what's *gained* by disabling this option.

> Other option can be a flag to enable automatic compilation per project, using a custom header like others used y Poedit.

This would even actually be more logical (especially given that some setups require the MO in a different location and differently named), yes.

> but a simple shortcut has a little cost too…

That may be, but it has *non-zero* cost. More things that may break (cross-platform, stepping over something), it eats up a shortcut that might be better used by something else in the future (this happened multiple times already), it signals that this menu item is *important* and should be used (it has, after all, a shortcut of its own), while in reality it's not so. I'm vary because I've *been* burned by haphazardly added shortcuts in the past.

By the way: note that on a Mac, you can add your own shortcuts to applications in *System Preferences → Keyboard → Shortcuts → App Shortcuts* yourself.

vaites:
Whoops, I swapped "PO" and "MO", you're right. Forgive me...

I disabled the automatic compilation for various completely subjective reasons:
- Keep only text files on translations folder (some projects not even need the MO file)
- Avoid compiled files on Git, like any other assets (minified frontend files, etc...)
- I only request PO files to external translators
- The MO file can be only compiled with build automation

[...]

PR #527, opened 2018-11-08
Add keyboard shortcut F8 to menu_compile_mo

c72578:
- Fixes https://github.com/vslavik/poedit/issues/526


vslavik:
After thinking about it, I don’t think this item needs a shortcut. Having an accelerator might make sense, but this is not a frequent action warranting a shortcut.

The other reason why I won’t be merging this PR, at least not in this form, is that F8 is just not a good _cross-platform_ one — it’s completely alien on Macs. I’m not sure where it comes from?

vslavik:
(Shift+Ctrl+C might be a better fit, also consistent with the historical HTML shortcut that too IMHO shouldn't exist...)

PR #533, opened 2018-11-26
Use remote part's implementation of ccache

Lin-Buo-Ren:
This patch uses the remote part implementation of ccache to shrink
the lines of code in the recipe.  The implementation of the remote
part can be acquired by running `snapcraft define ccache`.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -332,40 +332,6 @@ parts:
   ccache:
     after:
       - gcc-6
-    override-pull: 'true'
-    build-packages:
-      - ccache
-    plugin: nil
-    override-build: |
-      mkdir \
-        --parents \
-        "${SNAPCRAFT_PART_INSTALL}"/bin
-      ln \
-        --force \
-        --symbolic \
-        "$(which ccache)" \
-        "${SNAPCRAFT_PART_INSTALL}"/bin/gcc-6
-      ln \
-        --force \
-        --symbolic \
-        "$(which ccache)" \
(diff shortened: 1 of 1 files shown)


vslavik:
> The implementation of the remote part can be acquired by running snapcraft define ccache.

What does that mean? That more steps are required now?

snap/snapcraft.yaml
```diff
@@ -332,40 +332,6 @@ parts:
   ccache:
```
vslavik: I'm confused: if this removes `ccache`, why is this bit still here? And how does the snap know to use ccache when the dependency isn't listed anywhere?

Lin-Buo-Ren:
> What does that mean? That more steps are required now?

It simply means you can check/review the new implementation by running aforementioned command.

snap/snapcraft.yaml
Lin-Buo-Ren: Refer [Remote (reusable) parts - doc - snapcraft.io](https://forum.snapcraft.io/t/remote-reusable-parts/4233) for more info about remote parts.  Snapcraft maintains a regularly updated cache of remote parts.

vslavik:
So this part won't need `source: snap/local/empty` from #537 anymore either?

Lin-Buo-Ren:
Currently, [the ccache remote part builds CCache from upstream source](https://github.com/Lin-Buo-Ren/ccache-part/blob/8dcec4d0f6cb1dd2efacd01ab60d44785a8d8347/snapcraft.yaml#L17), as the part's `source` key is defined the `snap/local/empty` workaround isn't necessary.

vslavik:

PR #536, opened 2018-11-28
Move non-snapcraft files to snap/local

Lin-Buo-Ren:
Since snapcraft 3.0 and 2.44 the files that are not snapcraft-related
should reside in the snap/local directory, if not the following warning
message will appear:

```
The snap/ directory is meant specifically for snapcraft, but it contains
the following non-snapcraft-related paths, which is unsupported and will
cause unexpected behavior:
- .editorconfig
- README

If you must store these files within the snap/ directory, move them to
snap/local/, which is ignored by snapcraft.
```

This patch moves all files to the new subdir with the exception of those
mentioned above, as they can be only applied to the snap/ directory.  It
is left to the maintainer to decide to keep, move, or drop them.

Refer: [sanity checks: allow snap/local dir by kyrofa · Pull Request #2268 · snapcore/snapcraft](https://github.com/snapcore/snapcraft/pull/2268)


vslavik:

PR #537, opened 2018-11-28
Workaround snapcraft's implicit source 'feature'

Lin-Buo-Ren:
Snapcraft will implicitly set a part's `source` as `.`(the directory
running the `snapcraft` command) when its `source` key is not set,
this causes the following problems:

* The entire source tree being copied to the part's srcdir
  (which is previously workarounded using `override-pull` scriptlets)
* The parts that using the `nil` plugin(parts without a source) are
  considered 'changed' whenever any file in the source tree is modified
  , causing all depending parts cleaned and rebuild by snapcraft.

This patch workaround it differently by using the snap/local/empty dummy
folder as these parts' `source` thus avoid the problems.

### DIFF
--- a/snap/local/empty/README
+++ b/snap/local/empty/README
@@ -1,0 +1,23 @@
+This is an empty source directory to workaround a snapcraft 'feature' 
+where parts using the nil plugin are implicitly considered to use `.` 
+as an implicit source directory.
+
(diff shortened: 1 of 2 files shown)


vslavik:

PR #542, opened 2019-01-07
Update autoconf-archive macros

c72578:
Update macros used to upstream’s v2019.01.06 release.

- Fixes build issue of Poedit on powerpc64 and powerpc64le archs
  on some Linux distros. See upstream commit:
  https://github.com/autoconf-archive/autoconf-archive/commit/4b3c263


vslavik:
Thank you!

PR #544, opened 2019-01-16
Snapcraft: simplify $prefix/share files packaging

vslavik:
No reason to split share content into multiple entries (and accidentally omit some), just package the entire `share/` directory.

Motivation: b3b28e047b4ceb18943fca6ea6ba97b305fccb76, similarly appdata files are accidentally not packages (but probably not useful).

@Lin-Buo-Ren can you see any problem with this simplification, i.e. could it possibly pull in some junk from other parts? TIA!

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -165,30 +165,17 @@ parts:
         - bin/*
       files-in-stage-packages:
         - usr/*
-      freedesktop-desktop-entries:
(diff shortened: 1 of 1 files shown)


Lin-Buo-Ren:
The filesets rules only apply to the belonging part, I see no problem here.

vslavik:
Thanks!

PR #547, opened 2019-01-31
Fixed typo

Vitexus:


c72578:
Most common is the conventional Unixes, but Unices, treating Unix as a Latin noun of the third declension, is also popular.
From: https://en.wikipedia.org/wiki/Unix

vslavik:
Yep. "Unixes" would be a choice I'd use today, or during some larger rewrite of the text, but not much point in doing a separate commit just for the sake of changing this. Thanks for the PR though!

PR #572, opened 2019-05-16
Fix wxBitmap/wxIcon asserts

### LINKED ISSUE #571: "Wrong type, required: wxBitmap but: wxIcon" error displayed
## Environment

- Poedit version: 2.2.2
- wxGTK version: 3.0.4
- GTK+ version: 3.24.8
- Operating system: openSUSE Tumbleweed (20190512)

## Bug

"Wrong type, required: wxBitmap but: wxIcon"  error is displayed in a popup window.

It looks like commit b33970a replaced wxBitmap with wxIcon in PoeditListCtrl but a part of the code still assumes the presence of wxBitmap.

I tried [this naive patch](https://build.opensuse.org/package/view_file/home:1Antoine1:branches:M17N/poedit/poedit-2.2.2-null-variant.patch?expand=1) and it seems to fix the problem - at least I don't have the error anymore in my environment.

### Steps to reproduce

1. Open example po file: [yast-rpm-groups-master-fr.po.zip](https://github.com/vslavik/poedit/files/3185689/yast-rpm-groups-master-fr.po.zip)
2. Scroll until you reach the "C++" chain.

Problem occurs when a chain is displayed for which there is not icon in first column.

vslavik:
Fixes #571.

@super7ramp could you check if these two patches work for you. They are your changes (and so committed under your name) split into two with some changes for compat with wx > 3.1.

### DIFF
--- a/src/edlistctrl.cpp
+++ b/src/edlistctrl.cpp
@@ -230,13 +230,6 @@ PoeditListCtrl::Model::Model(TextDirection appTextDir, ColorScheme::Mode visualM
     m_iconBookmark = wxArtProvider::GetIcon("ItemBookmarkTemplate");
     m_iconError = wxArtProvider::GetIcon("StatusError");
     m_iconWarning = wxArtProvider::GetIcon("StatusWarning");
-
-#ifdef HAS_BROKEN_NULL_BITMAPS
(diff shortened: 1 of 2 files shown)


super7ramp:
Tested OK :thumbsup:

vslavik:
@super7ramp Thanks! v2.2.3 incoming.

PR #576, opened 2019-05-21
Add header for LearnMoreLink.

0-wiz-0:
Fixes
welcomescreen.cpp: In constructor 'EmptyPOScreenPanel::EmptyPOScreenPanel(PoeditFrame*)':
welcomescreen.cpp:289:26: error: 'LearnMoreLink' does not name a type; did you mean 'learnMore'?
     auto learnMore = new LearnMoreLink(this, "http://www.gnu.org/software/gettext/manual/html_node/", _("(Learn more about GNU gettext)"));
                          ^~~~~~~~~~~~~
                          learnMore

### DIFF
--- a/src/welcomescreen.cpp
+++ b/src/welcomescreen.cpp
@@ -26,6 +26,7 @@
 #include "welcomescreen.h"
 
 #include "colorscheme.h"
+#include "customcontrols.h"
 #include "crowdin_gui.h"
 #include "edapp.h"
 #include "edframe.h"


vslavik:
What compiler is this? Any idea why `configure` decided it doesn't support pre-compiled headers?

(Doesn't make the patch any less valid, of course, I'm just surprised that it caused a problem *in practice*.)

0-wiz-0:
This is on NetBSD 8.99.38/amd64 with the gcc coming with it, which is gcc (nb3 20190319) 7.4.0.
NetBSD's stance on precompiled headers with gcc is that they are broken - in particular, they don't play well with e.g. on 32bit systems or ASLR in general.

leio:
> What compiler is this? Any idea why `configure` decided it doesn't support pre-compiled headers?
> 
> (Doesn't make the patch any less valid, of course, I'm just surprised that it caused a problem _in practice_.)

There is a (to me) very obvious reason for that -- configure.ac appears to have zero checking for pre-compiled headers and Makefile.am's have zero usages of pre-compiled headers. It will simply always fail to compile with configure usage when building 2.2.3 without patching it, because there is no precompiled headers support in the first place - I'm not sure why that was asserted here at all.

vslavik:
> I'm not sure why that was asserted here at all.

Because people are fallible and make mistakes, you know. I'm so used to wx having PCH support and Poedit having it on all other platforms that the possibility that I wouldn't have added PCH support to autoconf builds simply didn't occur to me.

leio:
So that's a hint to add it, I guess :)
Given this, any plans for a 2.3.4 to fix this in releases? I grabbed the patch for now already though.

PR #583, opened 2019-06-09
Snapcraft: Migrate to base snap and modern syntax

Lin-Buo-Ren:
This patch implements the necessary changes to adopt the use of the
`core` base snap, which also makes Snapcraft no longer runs in legacy
2.43 mode and enables its new functionalities(such as Layout[1]).

* The empty source directory workaround is no longer required and is now
dropped
* Remote part feature is obsoleted thus all remote parts are now defined
locally(it is superseded by the new "extensions[2]" and "stage snaps[3]"
feature, which are not available in practice, yet)
* `no-system-libraries` build attribute has become the default behavior
and is obsoleted and thus dropped.

[1]: https://forum.snapcraft.io/t/snap-layouts/7207
[2]: https://forum.snapcraft.io/t/release-notes-snapcraft-3-0/10704#heading--extensions
[3]: https://forum.snapcraft.io/t/stage-snaps/11001


GitCop:

 * Commit: 56610258ab4ce725f09a317d070dc8197257e569

---

Lin-Buo-Ren:
Note that modern Snapcraft now building the snap in a fresh, minimal Ubuntu VM by default (requires [Multipass](https://github.com/CanonicalLtd/multipass), which is [available in Mac OS](https://discourse.ubuntu.com/t/installing-multipass-on-macos/8329)).  You may also fire up an LXD container instead by using the `--use-lxd` switch.

vslavik:
> * Remote part feature is obsoleted thus all remote parts are now defined
locally(it is superseded by the new "extensions[2]" and "stage snaps[3]"
feature, which are not available in practice, yet)

Wouldn't it be better to wait a bit, then, and do the upgrade cleanly without re-introducing cruft? (Sorry if it's a stupid question, I do't follow snapcraft development that closely...)

Lin-Buo-Ren:
Making the change now will allow us to subsequently:

* Drop the enchant, gtkspell, and iso-codes part definitions by using the layouts feature in the new syntax to solve the relocation problem of the iso-codes database
* Potentially migrate the snap to `core18` base (based on Ubuntu 18.04), which makes the gcc-6 part unnecessary as it should be available from the stage-packages

You may want to track the progress of [the gnome extension implementation](https://github.com/snapcore/snapcraft/pull/2398).

vslavik:
> Making the change now will allow us to subsequently:

To be clear, I'm not disputing the usefulness of this PR — it absolutely should be merged *eventually*. I'm just wondering if it shouldn't be postponed until it can be done *cleanly* (i.e. without reintroducing boilerplate that would temporarily make the definition file *worse*, not better). It seems to be that there are no immediate advantages to this upgrade *now*, are there? 

> Potentially migrate the snap to core18 base (based on Ubuntu 18.04), which makes the gcc-6 part unnecessary

As well as the `boost` and `cpprest` ones, plus getting more updated versions of the other stuff (`gettext` binaries being of primary concern here)…

Lin-Buo-Ren:
> I'm just wondering if it shouldn't be postponed until it can be done _cleanly_ (i.e. without reintroducing boilerplate that would temporarily make the definition file _worse_, not better).

I have no definite answer for that as the gnome extension implementation doesn't have an ETA.

> It seems to be that there are no immediate advantages to this upgrade now, are there?

The immediate advantage is that the follow-up improvements won't be blocked by this one.  I'll leave it to your discretion.

vslavik:
I think it would make most sense to have it on a dedicated branch where further consequent snap improvements can be made, then merge it all together once it provides clear improvements. This PR's branch would serve well for a series of thematically related commits, but if you prefer, I can pull it into a `snap3` branch in the upstream repo.

Lin-Buo-Ren:
> but if you prefer, I can pull it into a snap3 branch in the upstream repo

That would be preferable.

vslavik:
OK, [branch `snapcraft-3`](https://github.com/vslavik/poedit/tree/snapcraft-3) created, currently containing only this PR.

Thanks!

vslavik:
FYI, I'm going to merge this branch to `master` soon. Your work on this enabled me to get my bearings with snapcraft and upgrade Poedit's snap to core18 and gnome-3-28 extension, which among other things simplifies the build and addresses #591. So thanks again for doing it!

PR #590, opened 2019-07-28
Improve README

pereorga:


README
```diff
@@ -40,7 +40,7 @@ Requirements:
 
 Optional dependencies:
 
-  * cld2 (better language autodetection and non-English source languages)
```
vslavik: This change is semantically incorrect — it really is *and*, not *in*.

README
```diff
@@ -53,7 +53,7 @@ Do the usual thing:
     make
     make install
 
-You must have the dependencies installed in a location configure will find,
```
vslavik: I think "will" is better here, and a better fix would be "in a location *where* configure will find _them_,"

README
pereorga: Ok, I was unsure about this one (did not understand the meaning). I will update

README
pereorga: I see, your suggestion makes sense (it's the "dependencies", not the "location", that `configure` will find). I will update

pereorga:
Pushed. Also, removed a trailing space.

README
```diff
@@ -12,15 +12,15 @@
...
+with translating applications into another language. For details on principles
 of the solution used, see GNU gettext documentation or wxWidgets' wxLocale
 class reference.
 
 
  Installing
 ============
 
-Easily-installable prebuilt binaries are provided for Windows and macOS and
+Easily-installable prebuilt binaries are provided for Windows and macOS are
```
c72578: The second "are" in the sentence is not correct yet.
Easily-installable prebuilt binaries are provided for Windows and macOS **are** available from https://poedit.net/download.

How about a shortened version:
Easily-installable prebuilt binaries for Windows and macOS are available from https://poedit.net/download.

pereorga: Right, that makes sense

vslavik:
Thanks!

PR #599, opened 2019-08-09
README: Add Debian/Ubuntu compilation instructions

pereorga:
I had to install the following to get rid of most warnings when running `./configure`.

I guess `build-essential` package was also required.

### DIFF
--- a/README
+++ b/README
@@ -54,7 +54,14 @@ Do the usual thing:
     make install
 
 You must have the dependencies installed in a location where configure will find
-them, e.g. by setting CPPFLAGS and LDFLAGS appropriately.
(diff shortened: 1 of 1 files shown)


vslavik:
Thanks, appreciated! However, I thought about it and I'm going to pass on this one.

It's a slippery slope: first we add this and then the next gazillion of distros want their own unique instructions too (and be justified in the ask by this precedent).

On a slightly philosophical note, compiling from sources is never a trivial endeavor and it requires some minimal competence in dealing with issues that may show up; knowing basics about one's distro like how to install its compiler or how it handles dev packages is arguably part of that. In that regard, a little bit of friction is arguably *good* — it makes the person who wishes to compile from sources at least google "how to compile from sources on $DISTRO" and be better equipped to interpret results on their own.

Issue #384, opened 2017-04-14
Add an option to have more than one translation per string


ItielMaN:
This is a follow-up for the discussion in:
https://crowdin.com/translate/poedit/3/en-he#1311

Some strings can have double meaning*, and in some languages (Hebrew included) translating them in one way won't be accurate for other instances of the same string.

*Not necessarily. Strings can be translated differently while having the same meaning, but having different translation for the same string may be needed for other purposes. See the explanation in the link above for an example.

vslavik:
Quoting:

> I see there's only one occurrence of this string here, but it's being used more than once in Poedit.
Can you please add another "Save" entry for translation? One for the main UI ribbon and another one for all the others.
> The reason I'm asking this is because the "Save changes" dialog has 3 buttons, including "Save" and "Don't save" which are opposites to each other, and translating both of them in a consistent way makes the "Save" translation in the main UI look inconsistent with the rest of the strings in the same ribbon. Hope that makes sense.
>
> @vslavik I was afraid you'd ask me to get into details :(
In hebrew, there are differences in how you write (and say) verbs for male and female. Most hebrew translators, including myself, prefer to translate verbs (and sentences) as "genderlessly" as possible, when possible. So in this case instead of "שמור", we use the general form of the verb (שמירה) that can be used for both male and female (same for "Open"- פתיחה instead of פתח, "Update"- עדכון instead of עדכן and so on).
The downside of it is that you can't say something like "don't שמירה" (as in "Don't save" button in the "Save changes" dialog), so I have to revert back to "שמור", which means I'm either way breaking the consistency of the other translations in the same section.

liftarn:
Isn't this solved by using contexts? https://www.gnu.org/software/gettext/manual/html_node/Contexts.html

vslavik:
(requires `msgctxt` support in wx)

vslavik:
> (requires msgctxt support in wx)

Unblocked by https://github.com/wxWidgets/wxWidgets/pull/530

PR #624, opened 2020-03-07
Use wxGETTEXT_IN_CONTEXT() for "Needs work"

lanurmi:
Take advantage of wxWidgets 3.1.1's support for translation contexts
and use it for "Needs Work"/"Needs work".

This very string has been troublesome with Finnish, as the translation
would need to be different in the editing area and in statistics.

There's probably one more "Needs work" entry in the closed-source statistics page.

(This commit message is just a placeholder explaining what the commit does and why.
I am confident I couldn't write a commit message that satisfies you anyway, so feel free to
rephrase.)

### DIFF
--- a/locales/Makefile.am
+++ b/locales/Makefile.am
@@ -28,6 +28,7 @@ XARGS=xargs
 
 # common xgettext args: C++ syntax, use the specified macro names as markers
 XGETTEXT_ARGS=-C -k_ -kwxGetTranslation -kwxTRANSLATE -kwxPLURAL:1,2 -F -j \
+              -kwxGETTEXT_IN_CONTEXT:1c,2 -kwxGETTEXT_IN_CONTEXT_PLURAL:1c,2,3 \
               --add-comments=TRANSLATORS \
               --from-code=UTF-8 \
               --package-name=Poedit --package-version=$(PACKAGE_VERSION) \
--- a/scripts/refresh-pot.sh
+++ b/scripts/refresh-pot.sh
@@ -5,6 +5,7 @@ PACKAGE_VERSION=2.1
 [ -n "${WXRC}" ] || WXRC=wxrc
 
 XGETTEXT_ARGS="-C -k_ -kwxGetTranslation -kwxTRANSLATE -kwxPLURAL:1,2 -F \
+              -kwxGETTEXT_IN_CONTEXT:1c,2 -kwxGETTEXT_IN_CONTEXT_PLURAL:1c,2,3 \
               --add-comments=TRANSLATORS \
               --from-code=UTF-8 \
               --package-name=Poedit --package-version=${PACKAGE_VERSION} \
--- a/src/editing_area.cpp
+++ b/src/editing_area.cpp
@@ -453,7 +453,9 @@ void EditingArea::CreateEditControls(wxBoxSizer *sizer)
     // "needs review" implies that somebody else should review the string after
     // I am done with it (i.e. consider it good), while "needs work" implies I
     // need to return to it and finish the translation.
-    m_fuzzy = new SwitchButton(this, wxID_ANY, MSW_OR_OTHER(_("Needs work"), _("Needs Work")));
+    m_fuzzy = new SwitchButton(this, wxID_ANY,
+        MSW_OR_OTHER(wxGETTEXT_IN_CONTEXT("editing_area", "Needs work"),
+        wxGETTEXT_IN_CONTEXT("editing_area", "Needs Work")));
 #ifdef __WXOSX__
     m_fuzzy->SetWindowVariant(wxWINDOW_VARIANT_SMALL);
 #endif
--- a/src/manager.cpp
+++ b/src/manager.cpp
@@ -271,7 +271,7 @@ void ManagerFrame::UpdateListCat(int id)
     m_listCat->InsertColumn(0, _("Catalog"));
     m_listCat->InsertColumn(1, _("Total"));
     m_listCat->InsertColumn(2, _("Untrans"));
-    m_listCat->InsertColumn(3, _("Needs Work"));
+    m_listCat->InsertColumn(3, wxGETTEXT_IN_CONTEXT("manager", "Needs Work"));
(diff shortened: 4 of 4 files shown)


vslavik:
Thanks!

This addresses #384.

It would be nice if you could explain *why* the strings are ambiguous in Finnish, because that isn't obvious to non-Finns like me, and contexts like "editing_area" or "manager" don't add much in terms of contextual information. Is it verb vs X or something else? What if it gets added elsewhere, how to treat it then? What about *Edit → Translation needs work*, which one is it? Understanding this would help me with both proper contextualization in the future, and with grasping the severity.

Severity matters, because this isn't as simple as just applying the patch. Loosing existing translations is something I can handle at Crowdin, hopefully, but the larger problem is that wx-3.1.1 is unstable.  Until API-stable wx-3.2.0 is released, I can't make modern enough wx version a requirement on Unix. 

Which means that until then, the only way to do this is by defining own version of `wxGETTEXT_IN_CONTEXT` (an avenue towards that can be seen on line 1375 [here](https://github.com/wxWidgets/wxWidgets/pull/530/files#diff-9402b99a345f42d87ee10f6bf7a7c93bR1375)). That's worth doing for really bad stuff, but maybe isn't worth the hassle for some tiny nitpicking. I am incompetent to make that judgement call without more information, and would thus appreciate some informative help.

> (This commit message is just a placeholder explaining what the commit does and why.
I am confident I couldn't write a commit message that satisfies you anyway, so feel free to
rephrase.)

I can't say I appreciate such snide remarks in such inappropriate places. Feel free to vent as much as you want in the PR description & comments, but adding it into the commit message itself? That honestly feels like just being a jerk — knowing that *now* you've forced me to do extra work to rebase and edit what could otherwise be used as-is. 

As for the *subject* of the snideness, I always and consistently only ask that commit messages follow widely recognized, universally accepted and [amply documented](https://chris.beams.io/posts/git-commit/) git commit message guidelines that are thoughtfully explained and *backed by rational arguments* — those, as I suspect you know, boil down to some formatting and, indeed, "explaining what the commit does and why". You'll note that I hold *myself* to the same standard and the reason why I do is because I value quality and because I give a damn about the people who'll have to work with the code and git history *years* down the road.

lanurmi:
> It would be nice if you could explain _why_ the strings are ambiguous in Finnish, because that isn't obvious to non-Finns like me, and contexts like "editing_area" or "manager" don't add much in terms of contextual information.

Yes, I am not sure what the context is really meant to contain... The [gettext documentation](https://www.gnu.org/software/gettext/manual/html_node/Contexts.html) suggests it is something pretty short, and presumably it is not intended to be a substitute for the translator comments.

> Is it verb vs X or something else?

In Finnish, "needs work" conveniently translates with one word, an adjective. And here the ambiguity is about singular vs. plural, as in the editing area the meaning is "[this one] needs work", while e.g. in the statistics it's "[this many items whose status is] needs work". It is not a huge problem, the translation is just a little awkward in one place or another, depending on the choice between singular/plural. Other items in the statistics are in plural.

> What if it gets added elsewhere, how to treat it then? What about _Edit → Translation needs work_, which one is it?

If exactly the same string gets added elsewhere, then I suppose it would be wise to have it with a context. Assuming the menu item's exact string is only used in one place, then there is no ambiguity to solve.

> Severity matters, because this isn't as simple as just applying the patch. Until API-stable wx-3.2.0 is released, I can't make modern enough wx version a requirement on Unix.

Once I noticed that Poedit's git version uses 3.1 series, I kind of assumed using its new features is permitted, but I'm painfully aware of the practical difficulties of adopting an ABI-unstable 3.1 on Unix.

> Which means that until then, the only way to do this is by defining own version of `wxGETTEXT_IN_CONTEXT` 
> That's worth doing for really bad stuff, but maybe isn't worth the hassle for some tiny nitpicking.

Certainly not worth doing for just this one string in Finnish (and I don't recall there would be any other similar strings to worry about currently).

> > I am confident I couldn't write a commit message that satisfies you anyway, so feel free to
> > rephrase.)
> 
> I can't say I appreciate such snide remarks in such inappropriate places. Feel free to vent as much as you want in the PR description & comments, but adding it into the commit message itself? That honestly feels like just being a jerk — knowing that _now_ you've forced me to do extra work to rebase and edit what could otherwise be used as-is.

I know you are strict about commit messages following the guidelines, and in the past I have failed even when attempting to follow the guidelines you were referring to back then (probably [this one](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html). As a result, [here](https://github.com/vslavik/poedit/pull/76#issuecomment-42970312) you chose to publicly announce my failure, even though you apparently only added a newline to wrap the text (an obvious mistake on my side), and removed the trailing period of the summary line (which was not explicitly forbidden by the guideline at the time).

vslavik:
> Certainly not worth doing for just this one string in Finnish (and I don't recall there would be any other similar strings to worry about currently).

I realized there's a way to do it compatibly with wx-3.0: use context only for one of the strings, and fall back to context-less string with old wx versions.

Is my understanding correct, then, that "Needs work" in the main UI is singular, and the ones in both the manager (in your patch; column header) and in statistics (row header) are the same, plural?

lanurmi:
> I realized there's a way to do it compatibly with wx-3.0: use context only for one of the strings, and fall back to context-less string with old wx versions.
> 
> Is my understanding correct, then, that "Needs work" in the main UI is singular, and the ones in both the manager (in your patch; column header) and in statistics (row header) are the same, plural?

Yes, that is correct.

vslavik:
I committed the disambiguation in 95ae719, with you as the author — it is the simplified version discussed above, but still substantially your work.

Note that at Crowdin, the string didn't show up as new one due to peculiarities of what Crowdin sees as duplicate or not. It is editable there, but pre-filled with the identical translation. Had I known how Crowdin behaves before, I wouldn't have waited this long with committing the fix: I expected it to appear as a new string, requiring translations, and didn't want to break existing localizations in 2.4.x series. That turned out to be a non-issue and I could have merged it much sooner :(

Anyway: Thanks again for both the PR and the explanations!

PR #633, opened 2020-05-24
Crowdin API v2, allow editing of any files as XLIFF

tetatetit:
Added support (from customer's prospective):
- [Crowdin for enterprise](https://crowdin.com/enterprise)
  - Works with Crowdin Enterprise accounts of any organization (on any domains)
- Open/save/sync as Xliff any file which can be parsed for translations by Crowdin 
- While continue to work with PO natively as is

Improved
- Path to locally stored Crowdin files contains `<project name>/<lang code>/<branch><path to file in Crowdin project>`
- Files opened via `Open from Crowdin` can be saved/synced back to Crowdin on further open and edit as local file from file system (simplest from `File` menu's Recent list/history)

Migrated to new:
- OAuth improved/updated that works with API v2
- API v2

#### This PR is clean version of #629 with
- Work with PO natively (as is without converting to/from Xliff) is back
- Clean/arranged/grouped commits
- Back to implicit flows with few improvements
  - `state` param passed to OAuth URL and checked when callback received
  - OAuth URL redirect via poedit.net replaced with direct to Crowdin OAuth URL but source tracking params added (to identify authentication via Poedit for statistics)

### DIFF
--- a/src/catalog.cpp
+++ b/src/catalog.cpp
@@ -597,6 +597,9 @@ void Catalog::SetFileName(const wxString& fn)
 {
     wxFileName f(fn);
     f.Normalize(wxPATH_NORM_DOTS | wxPATH_NORM_ABSOLUTE);
+    auto name = f.GetName();
+    name.ToLong(&m_crowdinProjectId);
+    name.AfterFirst('_').ToLong(&m_crowdinFileId);
     m_fileName = f.GetFullPath();
 }
 
--- a/src/catalog.cpp
+++ b/src/catalog.cpp
@@ -602,8 +602,10 @@ void Catalog::SetFileName(const wxString& fn)
 
 bool Catalog::IsFromCrowdin() const
 {
-    if((m_crowdinFileId > 0 && m_crowdinProjectId) > 0
-        || (m_header.HasHeader("X-Crowdin-Project") && m_header.HasHeader("X-Crowdin-File")))
+    if (m_crowdinFileId > 0 && m_crowdinProjectId > 0)
+        return true;
+
+    if (m_header.HasHeader("X-Crowdin-Project") && m_header.HasHeader("X-Crowdin-File"))
         return true;
 
     auto name = wxFileName(m_fileName).GetName();
--- a/src/catalog.cpp
+++ b/src/catalog.cpp
@@ -597,12 +597,25 @@ void Catalog::SetFileName(const wxString& fn)
 {
     wxFileName f(fn);
     f.Normalize(wxPATH_NORM_DOTS | wxPATH_NORM_ABSOLUTE);
-    auto name = f.GetName();
-    name.ToLong(&m_crowdinProjectId);
-    name.AfterFirst('_').ToLong(&m_crowdinFileId);
     m_fileName = f.GetFullPath();
 }
 
+bool Catalog::IsFromCrowdin() const
+{
+    if((m_crowdinFileId > 0 && m_crowdinProjectId) > 0
+        || (m_header.HasHeader("X-Crowdin-Project") && m_header.HasHeader("X-Crowdin-File")))
+        return true;
+
+    auto name = wxFileName(m_fileName).GetName();
+    if(name.StartsWith("CrowdinSync_"))
+    {
+        name = name.AfterFirst('_');
+        name.ToLong(&m_crowdinProjectId);
+        name.AfterFirst('_').ToLong(&m_crowdinFileId);
+    }
+
+    return m_crowdinFileId > 0 && m_crowdinProjectId > 0;
+}
 
 namespace
 {
--- a/src/catalog.h
+++ b/src/catalog.h
@@ -469,7 +469,7 @@ class Catalog
         Type GetFileType() const { return m_fileType; }
 
         wxString GetFileName() const { return m_fileName; }
-        void SetFileName(const wxString& fn);
+        virtual void SetFileName(const wxString& fn);
 
         /**
             Return base path to source code for extraction, or empty string if not configured.
@@ -552,9 +552,10 @@ class Catalog
         virtual void SetLanguage(Language lang);
 
         /// Is the PO file from Crowdin, i.e. sync-able?
-        bool IsFromCrowdin() const
-            { return m_header.HasHeader("X-Crowdin-Project") && m_header.HasHeader("X-Crowdin-File"); }
-
+        bool IsFromCrowdin() const { return m_crowdinFileId > 0 && m_crowdinProjectId > 0; }
+        long GetCrowdinFileId() const { return m_crowdinFileId; }
+        long GetCrowdinProjectId() const { return m_crowdinProjectId; }
+            
         /// Returns true if the catalog contains obsolete entries (~.*)
         virtual bool HasDeletedItems() const = 0;
 
@@ -594,6 +595,7 @@ class Catalog
         bool m_isOk;
         Type m_fileType;
         wxString m_fileName;
+        long m_crowdinFileId = -1, m_crowdinProjectId = -1;
         HeaderData m_header;
         Language m_sourceLanguage;
 
--- a/src/catalog.h
+++ b/src/catalog.h
@@ -552,8 +552,7 @@ class Catalog
         virtual void SetLanguage(Language lang);
 
         /// Is the PO file from Crowdin, i.e. sync-able?
-        bool IsFromCrowdin() const { return (m_crowdinFileId > 0 && m_crowdinProjectId > 0) 
-                                            || (m_header.HasHeader("X-Crowdin-Project") && m_header.HasHeader("X-Crowdin-File")); }
+        bool IsFromCrowdin() const;
         long GetCrowdinFileId() const { return m_crowdinFileId; }
         long GetCrowdinProjectId() const { return m_crowdinProjectId; }
         void SetCrowdinProjectAndFileId(int projId, int fileId) { m_crowdinProjectId  = projId; m_crowdinFileId = fileId; }
@@ -597,7 +596,7 @@ class Catalog
         bool m_isOk;
         Type m_fileType;
         wxString m_fileName;
-        long m_crowdinFileId = -1, m_crowdinProjectId = -1;
+        mutable long m_crowdinFileId = -1, m_crowdinProjectId = -1;
         HeaderData m_header;
         Language m_sourceLanguage;
 
--- a/src/catalog.h
+++ b/src/catalog.h
@@ -469,7 +469,7 @@ class Catalog
         Type GetFileType() const { return m_fileType; }
 
         wxString GetFileName() const { return m_fileName; }
-        virtual void SetFileName(const wxString& fn);
+        void SetFileName(const wxString& fn);
 
         /**
             Return base path to source code for extraction, or empty string if not configured.
--- a/src/catalog.h
+++ b/src/catalog.h
@@ -552,9 +552,11 @@ class Catalog
         virtual void SetLanguage(Language lang);
 
         /// Is the PO file from Crowdin, i.e. sync-able?
-        bool IsFromCrowdin() const { return m_crowdinFileId > 0 && m_crowdinProjectId > 0; }
+        bool IsFromCrowdin() const { return (m_crowdinFileId > 0 && m_crowdinProjectId > 0) 
+                                            || (m_header.HasHeader("X-Crowdin-Project") && m_header.HasHeader("X-Crowdin-File")); }
         long GetCrowdinFileId() const { return m_crowdinFileId; }
         long GetCrowdinProjectId() const { return m_crowdinProjectId; }
+        void SetCrowdinProjectAndFileId(int projId, int fileId) { m_crowdinProjectId  = projId; m_crowdinFileId = fileId; }
             
         /// Returns true if the catalog contains obsolete entries (~.*)
         virtual bool HasDeletedItems() const = 0;
--- a/src/catalog_xliff.cpp
+++ b/src/catalog_xliff.cpp
@@ -29,6 +29,7 @@
 #include "configuration.h"
 #include "str_helpers.h"
 #include "utility.h"
+#include "crowdin_gui.h"
 
 #include <wx/intl.h>
 #include <wx/log.h>
@@ -378,7 +379,6 @@ bool XLIFFCatalog::CanLoadFile(const wxString& extension)
     return extension == "xlf" || extension == "xliff";
 }
 
-
 std::shared_ptr<XLIFFCatalog> XLIFFCatalog::Open(const wxString& filename)
 {
     constexpr auto parse_flags = parse_full | parse_ws_pcdata | parse_fragment;
@@ -403,6 +403,7 @@ std::shared_ptr<XLIFFCatalog> XLIFFCatalog::Open(const wxString& filename)
     else
         throw XLIFFReadException(filename, wxString::Format(_("unsupported XLIFF version (%s)"), xliff_version));
 
+    cat->SetFileName(filename);
     cat->Parse(xliff_root);
 
     return cat;
@@ -595,14 +596,23 @@ public:
     }
 };
 
-
 void XLIFF1Catalog::Parse(pugi::xml_node root)
 {
     int id = 0;
+    
     for (auto file: root.children("file"))
     {
-        m_sourceLanguage = Language::TryParse(file.attribute("source-language").value());
-        m_language = Language::TryParse(file.attribute("target-language").value());
+        // TODO: Only first `file` node atthributes are used for now
+        //       what works well in case of either all next are same within
+        //       same `xliff` or if the only single `file` node is there
+        //       (what is always the only case for downloaded from Crowdin)
+        //       But should be taken into account and improved for probable
+        //       more wide varietty of cases in the future.
+        if(id == 0) {
+            m_sourceLanguage = Language::TryParse(file.attribute("source-language").value());
+            SetLanguage(Language::TryParse(file.attribute("target-language").value()));
+        }
+        
         for (auto unit: file.select_nodes(".//trans-unit"))
         {
             auto node = unit.node();
@@ -737,7 +747,7 @@ protected:
 void XLIFF2Catalog::Parse(pugi::xml_node root)
 {
     m_sourceLanguage = Language::TryParse(root.attribute("srcLang").value());
-    m_language = Language::TryParse(root.attribute("trgLang").value());
+    SetLanguage(Language::TryParse(root.attribute("trgLang").value()));
 
     int id = 0;
     for (auto segment: root.select_nodes(".//segment"))
--- a/src/catalog_xliff.cpp
+++ b/src/catalog_xliff.cpp
@@ -608,7 +608,8 @@ void XLIFF1Catalog::Parse(pugi::xml_node root)
         //       (what is always the only case for downloaded from Crowdin)
         //       But should be taken into account and improved for probable
         //       more wide varietty of cases in the future.
-        if(id == 0) {
+        if(id == 0)
+        {
             m_sourceLanguage = Language::TryParse(file.attribute("source-language").value());
             SetLanguage(Language::TryParse(file.attribute("target-language").value()));
         }
--- a/src/catalog_xliff.cpp
+++ b/src/catalog_xliff.cpp
@@ -611,7 +611,7 @@ void XLIFF1Catalog::Parse(pugi::xml_node root)
         if(id == 0)
         {
             m_sourceLanguage = Language::TryParse(file.attribute("source-language").value());
-            SetLanguage(Language::TryParse(file.attribute("target-language").value()));
+            m_language = Language::TryParse(file.attribute("target-language").value());
         }
         
         for (auto unit: file.select_nodes(".//trans-unit"))
@@ -748,7 +748,7 @@ protected:
 void XLIFF2Catalog::Parse(pugi::xml_node root)
 {
     m_sourceLanguage = Language::TryParse(root.attribute("srcLang").value());
-    SetLanguage(Language::TryParse(root.attribute("trgLang").value()));
+    m_language = Language::TryParse(root.attribute("trgLang").value());
 
     int id = 0;
     for (auto segment: root.select_nodes(".//segment"))
--- a/src/catalog_xliff.h
+++ b/src/catalog_xliff.h
@@ -27,6 +27,7 @@
 #define Poedit_catalog_xliff_h
 
 #include "catalog.h"
+#include "cloud_sync.h"
 #include "errors.h"
 
 #include "pugixml.h"
@@ -101,9 +102,6 @@ public:
 
     ValidationResults Validate(bool wasJustLoaded) override;
 
-    Language GetLanguage() const override { return m_language; }
-    void SetLanguage(Language lang) override { m_language = lang; }
-
     // FIXME: PO specific
     bool HasDeletedItems() const override { return false;}
     void RemoveDeletedItems() override {}
@@ -118,7 +116,6 @@ protected:
 
 protected:
     pugi::xml_document m_doc;
-    Language m_language;
 };
 
 
--- a/src/catalog_xliff.h
+++ b/src/catalog_xliff.h
@@ -101,6 +101,9 @@ public:
 
     ValidationResults Validate(bool wasJustLoaded) override;
 
+    Language GetLanguage() const override { return m_language; }
+    void SetLanguage(Language lang) override { m_language = lang; }
+
     // FIXME: PO specific
     bool HasDeletedItems() const override { return false;}
     void RemoveDeletedItems() override {}
@@ -115,6 +118,7 @@ protected:
 
 protected:
     pugi::xml_document m_doc;
+    Language m_language;
 };
 
 
--- a/src/catalog_xliff.h
+++ b/src/catalog_xliff.h
@@ -27,7 +27,6 @@
 #define Poedit_catalog_xliff_h
 
 #include "catalog.h"
-#include "cloud_sync.h"
 #include "errors.h"
 
 #include "pugixml.h"
--- a/src/cloud_sync.h
+++ b/src/cloud_sync.h
@@ -83,6 +83,26 @@ public:
 
         return std::make_shared<Dest>(name, std::move(func));
     }
+
+    static const wxString& GetCacheDir()
+    {
+        static wxString localBaseDir;
+
+        if(localBaseDir.empty()) {
+
+            #if defined(__WXOSX__)
+                localBaseDir = wxGetHomeDir() + "/Library/Caches/net.poedit.Poedit";
+            #elif defined(__UNIX__)
+                if (!wxGetEnv("XDG_CACHE_HOME", &localBaseDir))
+                    localBaseDir = wxGetHomeDir() + "/.cache";
+                localBaseDir += "/poedit";
+            #else
+                localBaseDir = wxStandardPaths::Get().GetUserDataDir() + wxFILE_SEP_PATH + "Cache";
+            #endif
+        }
+
+        return localBaseDir;
+    }
 };
 
 
--- a/src/cloud_sync.h
+++ b/src/cloud_sync.h
@@ -89,7 +89,8 @@ public:
     {
         static wxString localBaseDir;
 
-        if(localBaseDir.empty()) {
+        if(localBaseDir.empty())
+        {
 
             #if defined(__WXOSX__)
                 localBaseDir = wxGetHomeDir() + "/Library/Caches/net.poedit.Poedit";
@@ -137,7 +138,8 @@ public:
     /// Show the window while performing background sync action. Show error if 
     static void RunSync(wxWindow *parent, std::shared_ptr<CloudSyncDestination> dest, CatalogPtr file)
     {
-        if(!dest->Auth(parent)) {
+        if(!dest->Auth(parent))
+        {
             return;
         }
 
--- a/src/cloud_sync.h
+++ b/src/cloud_sync.h
@@ -62,6 +62,7 @@ public:
 
     /// Asynchronously uploads the file. Returned future throws on error.
     virtual dispatch::future<void> Upload(CatalogPtr file) = 0;
+    virtual bool Auth(wxWindow* parent) = 0;
 
 
     /// Convenicence for creating a destination from a lambda.
@@ -136,6 +137,10 @@ public:
     /// Show the window while performing background sync action. Show error if 
     static void RunSync(wxWindow *parent, std::shared_ptr<CloudSyncDestination> dest, CatalogPtr file)
     {
+        if(!dest->Auth(parent)) {
+            return;
+        }
+
         wxWindowPtr<CloudSyncProgressWindow> progress(new CloudSyncProgressWindow(parent, dest));
 #ifdef __WXOSX__
         progress->ShowWindowModal();
--- a/src/cloud_sync.h
+++ b/src/cloud_sync.h
@@ -30,6 +30,7 @@
 #include "concurrency.h"
 
 #include <wx/string.h>
+#include <wx/stdpaths.h>
 
 #if wxUSE_GUI
     #include "customcontrols.h"
--- a/src/cloud_sync.h
+++ b/src/cloud_sync.h
@@ -31,6 +31,7 @@
 
 #include <wx/string.h>
 #include <wx/stdpaths.h>
+#include <wx/utils.h>
 
 #if wxUSE_GUI
     #include "customcontrols.h"
--- a/src/crowdin_client.cpp
+++ b/src/crowdin_client.cpp
@@ -44,7 +44,6 @@
 
 #include <iostream>
 
-
 // GCC's libstdc++ didn't have functional std::regex implementation until 4.9
 #if (defined(__GNUC__) && !defined(__clang__) && !wxCHECK_GCC_VERSION(4,9))
     #include <boost/regex.hpp>
@@ -256,7 +255,7 @@ dispatch::future<std::vector<CrowdinClient::ProjectListing>> CrowdinClient::GetU
 dispatch::future<CrowdinClient::ProjectInfo> CrowdinClient::GetProjectInfo(const int project_id)
 {
     auto url = "projects/" + std::to_string(project_id);
-    auto prj = make_shared<ProjectInfo>();
+    auto prj = std::make_shared<ProjectInfo>();
     enum { NO_ID = -1 };
         
     return m_api->get(url)
@@ -293,10 +292,10 @@ dispatch::future<CrowdinClient::ProjectInfo> CrowdinClient::GetProjectInfo(const
     {
         // Handle directories
         struct dir {
-            string name;
+            std::string name;
             int parentId;
         };
-        map<int, dir> dirs;
+        std::map<int, dir> dirs;
 
         for(const auto& i : r["data"]) {
             const json& d = i["data"],
@@ -310,7 +309,7 @@ dispatch::future<CrowdinClient::ProjectInfo> CrowdinClient::GetProjectInfo(const
             });
         }
 
-        stack<string> path;
+        std::stack<std::string> path;
         for(auto& i : prj->files) {
             int dirId = i.dirId;
             while(dirId != NO_ID) {
@@ -318,7 +317,7 @@ dispatch::future<CrowdinClient::ProjectInfo> CrowdinClient::GetProjectInfo(const
                 path.push(dir.name);
                 dirId = dir.parentId;
             }
-            string pathStr;
+            std::string pathStr;
             while(path.size()) {
                 pathStr += '/';
                 pathStr += path.top();
@@ -334,11 +333,11 @@ dispatch::future<CrowdinClient::ProjectInfo> CrowdinClient::GetProjectInfo(const
     }).then([this, url, prj](json r)
     {
         // Handle branches
-        map<int, string> branches;
+        std::map<int, std::string> branches;
 
         for(const auto& i : r["data"]) {
             const json& d = i["data"];
-            branches[d["id"]] = d["name"].get<string>();
+            branches[d["id"]] = d["name"].get<std::string>();
         }
 
         for(auto& i : prj->files) {
--- a/src/crowdin_client.cpp
+++ b/src/crowdin_client.cpp
@@ -196,20 +196,15 @@ dispatch::future<CrowdinClient::UserInfo> CrowdinClient::GetUserInfo()
             const json& d = r["data"];
             UserInfo u;
             u.login = str::to_wstring(d["username"]);
-            u.name = str::to_wstring(d.value("fullName", ""));
-            if(u.name.empty())
+            try { u.name = str::to_wstring(d.at("fullName")); }
+            catch(...)
             {
-                // Take care that both first and last name can be empty
-                auto name = d["firstName"];
-                if(name.is_string())
-                    u.name += str::to_wstring(name);
-                if(!u.name.empty())
-                    u.name += ' ';
-                name = d["lastName"];
-                if(name.is_string())
-                    u.name += str::to_wstring(name);
+                std::string firstName, lastName;
+                try { firstName = d.at("firstName"); } catch(...) {}
+                try { lastName = d.at("lastName"); } catch(...) {}
+                u.name = str::to_wstring(firstName + ' ' + lastName);
             }
-            if(u.name.empty())
+            if(u.name.empty() || u.name == L" ")
                 u.name = u.login;
             return u;
         });
--- a/src/crowdin_client.cpp
+++ b/src/crowdin_client.cpp
@@ -229,15 +229,21 @@ dispatch::future<std::vector<CrowdinClient::ProjectListing>> CrowdinClient::GetU
             for (auto&& d : r["data"])
             {
                 const json& i = d["data"];
-                all.push_back({
-                    str::to_wstring(i["name"]),
-                    i["id"].get<int>(),
-                    //TODO: should be tested with project not owning by
-                    //      currently authorized user as well (not only
-                    //      for owning) after it will be implemented in
-                    //      in API v2 (unimplemented yet)
-                    /*(bool)i["publicDownloads"].get<int>()*/true
+                auto itPublicDownloads = i.find("publicDownloads");
+                if(itPublicDownloads != i.end()
+                    // for some wierd reason `publicDownloads` can be in 3 states:
+                    // `null`, `true` and `false` and, as investigated experimentally,
+                    // only `null` means 'Forbidden' to work with project files (to get list etc)
+                    // so hide such projects. While `false` or `true` allows to work with 
+                    // such projects normally
+                    && !itPublicDownloads->is_null()) {
+
+                    all.push_back({
+                        str::to_wstring(i["name"]),
+                        i["id"].get<int>()
                 });
+
+                }
             }
             return all;
         });
--- a/src/crowdin_client.cpp
+++ b/src/crowdin_client.cpp
@@ -100,12 +100,24 @@ protected:
     std::string parse_json_error(const json& response) const override
     {
         cout << "\n\nJSON error: " << response << "\n\n";
-        std::string msg = response["error"]["message"];
 
-        // Translate commonly encountered messages:
-        if (msg == "Translations download is forbidden by project owner")
-            msg = _("Downloading translations is disabled in this project.").utf8_str();
-        return msg;
+        try
+        {
+            return response.at("errors").at(0).at("error").at("errors").at(0).at("message");
+        }
+        catch(...)
+        {
+            try
+            {
+                return response.at("error").at("message");
+            }
+            catch (...)
+            {
+                std::string msg;
+                msg = _("JSON request error").utf8_str();
+                return msg;
+            }
+        }
     }
 
     void on_error_response(int& statusCode, std::string& message) override
--- a/src/crowdin_client.cpp
+++ b/src/crowdin_client.cpp
@@ -37,6 +37,10 @@
 #include <wx/translation.h>
 #include <wx/utils.h>
 
+#include <iostream>
+
+using namespace std;
+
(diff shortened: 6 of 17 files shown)


src/crowdin_gui.h
```diff
@@ -83,7 +83,7 @@ class CrowdinSyncDestination : public CloudSyncDestination
 {
 public:
     wxString GetName() const override { return "Crowdin"; }
-
+    bool Auth(wxWindow* parent) override;
```
vslavik, line 86: I don't understand why this is done. It is committed as part of "Crashed on open cached Crowdin file if logged out", but none of it seems to be related to crashing anything, and surely crashes can be avoided in a simpler way?

There's also a commit that seems related with a message "Move AttachCloudSync() to GUI code" which *does no such thing at all* and which perhaps seems be describing *this* commit, but in that case again, I don't understand why is it needed.

src/crowdin_gui.cpp
```diff
@@ -602,20 +604,33 @@ void CrowdinSyncFile(wxWindow *parent, std::shared_ptr<Catalog> catalog,
...
                 auto tmpdir = std::make_shared<TempDirectory>();
-                auto outfile = tmpdir->CreateFileName("crowdin.po");
+                auto outfile = tmpdir->CreateFileName("crowdin." + filename.GetExt());
 
                 dispatch::on_main([=]{
                     dlg->Activity->Start(_(L"Downloading latest translations…"));
                 });
-
+                if(filename.GetExt().CmpNoCase("po"))// if not PO
+                    filename.SetFullName(filename.GetName());// set original filename extension
```
vslavik, line 628: How does this line "set original filename extension" and what is "original" extension?

Please use explicit '!= 0' tests in cmp functions, so that casual reader doesn't read them as returning boolean, or use wxString::IsSameAs() instead.

src/crowdin_gui.cpp
```diff
@@ -538,6 +516,35 @@ class CrowdinOpenDialog : public wxDialog
...
+        auto projIt = find_if(projects.begin(), projects.end(),
+                                [=](const auto& proj) { return projName == proj.identifier; });
+        if(projIt == projects.end())
+            throw Exception(_("File project not found for currently authorized user"));
+        return CrowdinClient::Get().GetProjectInfo(projIt->id).then([=](CrowdinClient::ProjectInfo proj)
+        {
+            auto fileIt = find_if(proj.files.begin(), proj.files.end(),
+                                [=](const auto& file) { return filePathName == file.pathName; });
+            if(fileIt == proj.files.end())
+                throw Exception(_("File path not found in it project in currently authorized user"));
```
vslavik, line 541: The two messages have inconsistent phrasing ("in user", "for user"), lack punctuation and aren't very clear (what is "file project"?).

It would be better to throw a single, same error text, which explains to the user the error is Crowdin related and is because the user is not authorized to access the file. For example:

"You are not authorized to sync this file to Crowdin."

Details of why (project or file ID mismatch, error obtained etc.) are an appropriate use of `wxLogTrace`.

src/crowdin_gui.cpp
```diff
@@ -496,33 +484,23 @@ class CrowdinOpenDialog : public wxDialog
...
+        wxFileName crowdinFileName(name, wxPATH_UNIX),
+                   localFileName =
+            CloudSyncDestination::GetCacheDir() + wxFILE_SEP_PATH + "Crowdin"
+            + wxFILE_SEP_PATH + projectName
+            + wxFILE_SEP_PATH + lang.Code()
+            + wxFILE_SEP_PATH + crowdinFileName.GetPath() + wxFILE_SEP_PATH
+            << projectId << '_' << fileId << '_' << crowdinFileName.GetFullName();
+       
+        if(localFileName.GetExt().CmpNoCase("po"))// if not PO
+            localFileName.SetFullName(localFileName.GetFullName() + ".xliff");
```
vslavik, line 498: So, storing XLIFF files as `foo.xliff.xliff`?

src/crowdin_gui.cpp
```diff
@@ -385,20 +386,13 @@ class CrowdinOpenDialog : public wxDialog
     {
         m_info = prj;
         // Put supported files first in the list:
-        std::vector<std::wstring> f_unsup;
+        std::vector<CrowdinClient::FileInfo> f_unsup;
```
vslavik, line 389: All files are supported now. You removed all code adding anything to `f_unsup`. Why keep the variable?

src/crowdin_client.cpp
```diff
@@ -143,139 +158,331 @@ dispatch::future<void> CrowdinClient::Authenticate()
...
+    std::string domain;
+    try
+    {
+        domain = json::parse(
+            base64_decode_json_part(std::string(
+                wxString(token).AfterFirst('.').BeforeFirst('.').utf8_str()
+        ))).at("domain").get<std::string>();
+        domain += ".";
+    }
+    catch(...) {}
```
vslavik, line 476: Don't you think this situation - unexpectedly bad token - warrants at least trace logging, given how logging-heavy all the rest of the changes are? An explanatory comment along the lines of "assume token is non-enterprise, fall through with empty domain" would help in understanding that it is intentional and *correct* that the error is ignored.

src/crowdin_client.cpp
```diff
@@ -143,139 +158,331 @@ dispatch::future<void> CrowdinClient::Authenticate()
...
 
 
-dispatch::future<void> CrowdinClient::UploadFile(const std::string& project_id,
-                                                 const std::wstring& file,
+dispatch::future<void> CrowdinClient::UploadFile(const long project_id,
                                                  const Language& lang,
+                                                 const long file_id,
+                                                 const std::string& file_extension,
                                                  const std::string& file_content)
 {
```
vslavik, line 386: The 'const' for numerical arguments is superfluous (here and elsewhere: numbers are passed by value in C++; const& is needed when passing by reference to avoid modification).

src/crowdin_client.cpp
```diff
@@ -143,139 +158,331 @@ dispatch::future<void> CrowdinClient::Authenticate()
...
+            // for XLIFF and PO files should be exported "as is" so set to `false`
+            { "exportAsXliff", !(ext == "xliff" || ext == "xlf" || ext == "po") }
+        }))
+        .then([this, output_file] (json r) {
+            std::string url = r["data"]["url"];
+            wxURI uri(url);
+            // Per download (local) client must be created since different domain
+            // per request is not allowed by HTTP client backend on some platorms.
+            // (e.g. on Linux).
+            auto downloader = std::make_shared<crowdin_http_client>(*this, std::string((uri.GetScheme() + "://" + uri.GetServer()).mb_str()));
```
vslavik, line 371: Don't use `mb_str()`. Where you can, use `str_helpers.h` for conversions for clarify; here that would be `str::to_utf8()`.

`mb_str()` would do unexpected things, it is locale dependent.

src/crowdin_client.cpp
```diff
@@ -143,139 +158,331 @@ dispatch::future<void> CrowdinClient::Authenticate()
...
+
+        for(const auto& i : r["data"])
+        {
+            const json& d = i["data"];
+            branches[d["id"]] = d["name"].get<std::string>();
+        }
+
+        for(auto& i : prj->files)
+            if(i.branchId != NO_ID)
+                i.pathName = "/" + str::to_wstring(branches[i.branchId]) + i.pathName;
```
vslavik, line 342: You're using wstring, if you concatenate it with a string literal, it should be a *wide* string literal, i.e. `L"/"`.

src/crowdin_client.cpp
```diff
@@ -50,64 +57,71 @@
...
     {}
 
 protected:
     std::string parse_json_error(const json& response) const override
     {
-        std::string msg = response["error"]["message"];
-
-        // Translate commonly encountered messages:
-        if (msg == "Translations download is forbidden by project owner")
-            msg = _("Downloading translations is disabled in this project.").utf8_str();
```
vslavik, line 108: Does the removal of this mean that this situation can no longer happen in V2 of the API?

src/crowdin_client.cpp
```diff
@@ -50,64 +57,71 @@
...
+            {
+                return response.at("error").at("message");
+            }
+            catch (...)
+            {
+                std::string msg;
+                msg = _("JSON request error").utf8_str();
+                return msg;
+            }
+        }
```
vslavik, line 124: Can you please include a comment either showing expected JSON structure of the reply, or linking to API docs showing it? I still find this throw-and-catch cascade opaque and hard to understand, a comment showing the data it works with would help.

In particular I am confused by there being more `errors` nested inside a singular `error` which is itself contained in a possible list of errors… Does the Crowdin API really return *list of errors that themselves contain multitude of errors*? 😳

And if multiple errors can be returned, shouldn't they be shown to the user? 

Or use the [contains()](https://github.com/nlohmann/json/pull/1474) function without throwing, but I can see how the cascade is actually less complicated to write...

src/cloud_sync.h
```diff
@@ -62,6 +64,7 @@ class CloudSyncDestination
 
     /// Asynchronously uploads the file. Returned future throws on error.
     virtual dispatch::future<void> Upload(CatalogPtr file) = 0;
+    virtual bool Auth(wxWindow* parent) = 0;
```
vslavik, line 67: You are making this API depend on GUI code for reasons that aren't entirely clear to me (see the other comment; but this place shows why it bothers me).

src/catalog_xliff.cpp
```diff
@@ -595,14 +596,24 @@ class XLIFF10CatalogItem : public XLIFF12CatalogItem
...
+        // TODO: Only first `file` node atthributes are used for now
+        //       what works well in case of either all next are same within
+        //       same `xliff` or if the only single `file` node is there
+        //       (what is always the only case for downloaded from Crowdin)
+        //       But should be taken into account and improved for probable
+        //       more wide varietty of cases in the future.
+        if(id == 0)
+        {
+            m_sourceLanguage = Language::TryParse(file.attribute("source-language").value());
+            SetLanguage(Language::TryParse(file.attribute("target-language").value()));
```
vslavik, line 614: I don't think any of these leftover SetLanguage calls are actually needed anymore?

src/crowdin_client.cpp
```diff
@@ -50,64 +57,71 @@
...
 
 namespace
 {
 
-#define OAUTH_CLIENT_ID     "poedit-y1aBMmDbm3s164dtJ4Ur150e2"
-#define OAUTH_AUTHORIZE_URL "/oauth2/authorize?response_type=token&client_id=" OAUTH_CLIENT_ID
+#define OAUTH_SCOPE         "project"
+#define OAUTH_CLIENT_ID     "k0uFz5HYQh0VzWgZmOpA"
+//      any arbitrary unique unguessable string (e.g. UUID in hex)
+#define OAUTH_STATE         "948cf13ffffb47119d6cfa2b68898f67"
```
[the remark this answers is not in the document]
tetatetit: Addressed 122ae7f02299c34deb14b28d592c500efe96d98e so now `state` is generated once per authentication

src/crowdin_client.cpp
```diff
@@ -50,64 +57,71 @@
...
-
 std::string CrowdinClient::WrapLink(const std::string& page)
 {
-    std::string url("https://poedit.net/crowdin");
+    std::string url("https://accounts.crowdin.com");
     if (!page.empty() && page != "/")
-        url += "?u=" + http_client::url_encode(page);
+        url += page;
     return url;
 }
```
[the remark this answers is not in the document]
tetatetit: Addressed 122ae7f . Also rearranged/refactored OAuth URL related code to get rid of currently needless overcomplications with `WrapLink()` since now we don't use redirect via [poedit.net](htps://poedit.net) or anything else that really needs to `WrapLink()` resulting compounding of URL from different parts from different places of code...unless I miss something so please correct me if otherwise. But honestly speaking I really experienced some  complications when tried to follow `WrapLink()` first time (it was one of my first acquaintance with Poedit source code) which I found unjustified for so simple purpose or use case. Absolutely no blame on you but just recalling your care for anybody new who might start to work on Poedit in the future, so here is humble opinion and feedback of such person here :) (since that time it was pretty new for me :) ).

src/http_client_casablanca.cpp
```diff
@@ -217,19 +217,29 @@ class http_client::impl
...
         http::http_request req(http::methods::POST);
-        req.headers().add(http::header_names::accept,     L"application/json");
+        for(const auto& h : hdrs)
+            req.headers().add(
+                // Take care about conversion from std::string in case of
+                // std::wstring used in headers as well what is default
+                // for _WIN32 in cpprest for some reason
+                // (it's safe since we expect only ANSI in header names)
+                utility::string_t(h.first.begin(), h.first.end()),
+                utility::string_t(h.second.begin(), h.second.end())
```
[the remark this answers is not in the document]
tetatetit: Addressed c47410c91534f46c737dc5f117d17dcba85adc5e

src/crowdin_client.cpp
```diff
@@ -200,8 +200,8 @@ dispatch::future<CrowdinClient::UserInfo> CrowdinClient::GetUserInfo()
             catch(...)
             {
                 std::string firstName, lastName;
-                try { firstName = d.at("firstName"); } catch(...) {}
-                try { lastName = d.at("lastName"); } catch(...) {}
+                try { firstName = d.at("firstName").get<std::string>(); } catch(...) {}
+                try { lastName = d.at("lastName").get<std::string>(); } catch(...) {}
```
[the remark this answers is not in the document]
tetatetit: > The commit's message is misleading as this is not about wxStandardPaths (and such fixups are best handled by fixing up the bad topical commit, but never mind)

Yep, I see there some changes unrelated to this commit intention at all leaked in even after arranging commits to make them more clean and clear in this PR. e.g. [like this one](https://github.com/vslavik/poedit/pull/633/files/e06429f3e0217d757a81ed0e19071e82e35570e7..98ab380263e281e8c71f281208bd22fa0140990f#diff-a1865a3025926eb82afbd7f4d9f13518). Also description remained inaccurate (wxStandardPaths() fix was made in some another commit or just was not needed any more). Though trying to avoid rebases in this final PR already unless really would make more convenient for you to continue review/discussion by PR, since after rebase GitHub usually loses references on places in source code in PR review saying that "can not found them due of most probable rebased" ... unless I just miss some trick to go around of this. So I hope you still are able to catch/follow purpose of changes in this commit or otherwise I will be happy to clarify everything you will ask/comment. 

> I think what you meant by "Buildable on Windows" was "Buildable with Visual Studio 16.5" and the correct fix for it landed on master in 03ba307, rendering most of the changes in the commit unneeded.

Addressed 2c5a6b710e97936fc251d114b443ce9eac00b79d

src/crowdin_client.cpp
tetatetit: Addressed b69f4e1c047ddea3d84d135329d655bda9caf9d9

src/crowdin_client.cpp
tetatetit: It might, but so far I realized it might happen by at least more than one different reasons so displaying message from server (i.e. in response of API request) as is usually more informative here. As well as there might be some other reasons existing now or added in the future we are just not aware about so we might just to miss to catch all of them in any case if to try to generalize

src/crowdin_client.cpp
tetatetit: Commented 278a08250a71def13b71847c9e962309d829e90e

> In particular I am confused by there being more errors nested inside a singular error which is itself contained in a possible list of errors… Does the Crowdin API really return list of errors that themselves contain multitude of errors?

I don't know what how to explain/formalize this and what the purpose purpose is it's done in such way. But just working with handling the fact :). 

> And if multiple errors can be returned, shouldn't they be shown to the user?

I never encountered it, at least while looked into responses with such structure returned by all API requests used in Poedit. But you think it's so important to implement displaying of "all errors" just for that rare potential or may be even pure hypothetical case they will be there - please unresolve this comment so I will implement or at least check with Crowdin API team whether it's really the case worth implementing

src/catalog_xliff.cpp
tetatetit: Why if I see e.g. below in files downloaded from Crowdin?
```
<file id="25" original="/Test word file.docx" source-language="en" target-language="uk" datatype="plaintext">
```

vslavik: > Why if I see e.g. below in XLiff files downloaded from Crowdin?

I don't understand what you mean by that. I was referring to the - unneeded as far as I can tell - change from the original `m_language = Language::TryParse(file.attribute("target-language").value());` to this.

src/crowdin_client.cpp
vslavik: > But you think it's so important to implement displaying of "all errors"

Nah, I was just trying to understand why what seems like convoluted extraction of errors takes place… Let's shrug it off as c'est la vie...

src/crowdin_client.cpp
vslavik: Displaying server's error is of course best. But I think you are misinterpreting this code: it doesn't hide server error, it replaces *one exact specific text*, with another. The reason was that

(a) in V1 api, that the error given by the server was cryptic (people didn't understand that "forbidden" is something they just need to enable in the settings and interpreted it as being forbidden from doing it by Crowdin or by Poedit's free version etc.); and

(b) this code ensure *translation* of by far the most common error message. I don't think Crowdin localizes their errors per `Accept-Language` but maybe they do now…

In any case, I can easily check if this error is still produced by V2 or not.

src/crowdin_client.cpp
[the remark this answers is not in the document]
vslavik: `WrapLink` is certainly an unfortunate name and the comment doesn't do a good job explaining it; I should renamed it to something like `CrowdinClient::CreateLinkToPage`.

But that doesn't mean it is not needed, and if anything, your PR shows the need for it: when changing the way Crowdin is linked, it is better to do it in one place, not in several places; in that regard, hardcoding UTM tracking is a regression. But not worries, I'll address it myself.

src/cloud_sync.h
tetatetit: - That's description of [Crashed on open cached Crowdin file if logged out](68d5e5e54fe019ee877b40a49f938edefa8acdc5) commit (where this change is introduced) should be "Crashed on **saving of** cached Crowdin file if logged out" instead, my bad sorry.  I've implemented it similarly with how it's done in [here in `CrowdinOpenFile()`](https://github.com/tetatetit/poedit/blob/278a08250a71def13b71847c9e962309d829e90e/src/crowdin_gui.cpp#L552) and [here in `CrowdinSyncFile()`]( https://github.com/tetatetit/poedit/blob/278a08250a71def13b71847c9e962309d829e90e/src/crowdin_gui.cpp#L575)
1. But I could not put `CrowdinLoginDialog` as well as any other Crowdin specific code directly to closest using it GUI code  [`RunSync()`](https://github.com/tetatetit/poedit/blob/278a08250a71def13b71847c9e962309d829e90e/src/cloud_sync.h#L141) which assumed to be abstract/generalized for any `CloudSyncDestination` so knowing nothing about `CrowdinSyncDestination`
2. The only alternative I seen to avoid GUI code in `CrowdinSyncDestination` is to make `Auth()` to just `return CrowdinClient::Get().IsSignedIn()` silently without displaying `CrowdinLoginDialog` if not authorized. But in this case Poedit would just silently save such file locally, without bringing user attention that the file can be saved (uploaded) to Crowdin as well as providing option to sign in to Crowdin immediately "in place".

But please feel free to "Unresolve" this conversation if you see above variant 2 or anything else as better alternative.

src/catalog_xliff.cpp
tetatetit: Ah, now I got what you meant, sorry for being unapprehensive. This was part of some issue I experienced when implemented uploading files directly as XLIFF with new v2 API but unfortunately and my bad I have not commented which exactly issues it was. But do you see/suspect/feel it might create any side effect? Since I see [`XLIFF2Catalog::Parse()` uses it as well](https://github.com/tetatetit/poedit/blob/278a08250a71def13b71847c9e962309d829e90e/src/catalog_xliff.cpp#L751)

src/crowdin_client.cpp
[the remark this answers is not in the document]
tetatetit: Could you please look how it's addressed now in rebased and force-pushed above mentioned commit a046666ddfea90b2dec190b3a204a12e4fc699f6 in response on your below messages (which disappeared from GitHub after rebasing and force-pushed above commit)
> Doing this successfully broke the "Learn more about Crowdin" link — while it previously went to crowdin.com, a page where one can learn what Crowdin is and what it has to offer, it now goes directly to a nondescript login page.

> Let's just restore WrapLink to how it was before this PR and call it a day?

I would highly recommend against this due of reasons I already mentioned in my previous message by this conversation. Plus in response to your more resent messages I would not mix usage of same function call it `CrowdinClient::CreateLinkToPage()` or whatever else since now base URL for OAuth and used to "Learn more about Crowdin" are different now, i.e. pointing on different unrelated resources which just lead to even more confusion when using single function for them. Also sorry if I missed some of important points in your explanation but I still don't get necessity of such single common function now when there are clearly separated and quite straightforward flow/logic/code/place for generation of URL for OAuth and absolutely unrelated link for URL "Learn more about Crowdin". 

>  it is better to do it in one place, not in several places; in that regard

But currently all of them are in `CrowdinClient` (`crowdin_client.h/cpp`) very closely each from other in the code (2 lines line by line in [`crowdin_client.h`](https://github.com/tetatetit/poedit/blob/a6788fa535fefc63a8957424a8dfb50a6796d588/src/crowdin_client.h#L51) and just in 2 logical places in `crowdin_client.cpp`  - [one at the beginning of file for macros](https://github.com/tetatetit/poedit/blob/a6788fa535fefc63a8957424a8dfb50a6796d588/src/crowdin_client.cpp#L72) and [another in `CrowdinClient::Authenticate()` ](https://github.com/tetatetit/poedit/blob/a6788fa535fefc63a8957424a8dfb50a6796d588/src/crowdin_client.cpp#L141)

> in that regard, hardcoding UTM tracking is a regression. But not worries, I'll address it myself.

Thanks a lot for taking care for this, but I am still not sure I understand what you mean here so I would higly appreciate if you would clarify (just to get what I understand wrong to learn lesson for future or so). I have put UTM to the separate macro in above 1st place of code and use it consistently in the 2nd place above. So what is wrong I did with it?

src/crowdin_client.cpp
tetatetit: I just like this `const` C++ feature exactly here not to prevent modifying of external (outside/input) variable passed as argument to function (I am well aware about passing by `value` vs `reference`) but _to avoid just occasional unattended modifying of such arguments **inside function locally**_ (with e.g `=` or passing it as ref to some other function). It makes me more calm and sure (and BTW Rust uses such approach by default out of the box) about local code inside function since add one more compile-time check to avoid at least fool-proof and just occasional (by distraction) changes. While it does not cost too much redundancy/verbosity to the code. But anyway if you dislike it - please "Unresolve" this conversation so I will remove it :) and please feel free to do not spend time on explanation if you lack time (although would be just interesting and higly appreciated for me personally to hear your point of view on this here just to get some lesson/knowledge exchange here).

src/catalog.h
```diff
@@ -469,7 +469,7 @@ class Catalog
         Type GetFileType() const { return m_fileType; }
 
         wxString GetFileName() const { return m_fileName; }
-        void SetFileName(const wxString& fn);
+        virtual void SetFileName(const wxString& fn);
```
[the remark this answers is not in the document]
tetatetit: Addressed 11f638e314af0b00dc312fae49ff591e5a373d1a

src/catalog_xliff.h
```diff
@@ -27,6 +27,7 @@
 #define Poedit_catalog_xliff_h
 
 #include "catalog.h"
+#include "cloud_sync.h"
```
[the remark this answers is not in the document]
tetatetit: Addressed d2ea08f5c768c92add5b9cae972db32f22de5ffd

src/http_client.cpp
```diff
@@ -34,7 +34,6 @@
 #include <boost/uuid/uuid_io.hpp>
 #include <boost/uuid/uuid_generators.hpp>
 
-
```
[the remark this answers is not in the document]
tetatetit: Addressed 731c2664557736a4d331ced306cebdbb285c1ca9 , sorry for occasinally removed/changed it.

src/catalog_xliff.cpp
vslavik: > Since I see XLIFF2Catalog::Parse() uses it as well

Because you changed it there too… The comment applies to all such uses.

I had the grep the sources on `master` w.r.t. SetLanguage() and it modifies the XML tree. I.e. it is for external use (e.g. when the *user* changes the language) of the class, while these uses are internal during parsing.

tetatetit: Sorry about confusion with that function in ` XLIFF2Catalog::Parse() ` as well. Really have forgotten about this due of lot of changes (now I see my mistake that really needed to make smaller linked PRs from the beginning of my work which would be done not against master bug against e.g. branch called `v2.4` or `crowdin-v2-api` as base. Anyway do you see any possible side effect from using `SetLanguage()` internally from inside `Parse()`. Since I definitely could not change `SetLanguage()` in 2 places just for fun :) so there definitely should be some reason to this, at least as it seemed for me. So I just afraid that to do not break any Xlifff<=>Crowdin functionality by reverting it back. That is why I am paying attention to this issue and asking you so thoroughly.

src/edframe.cpp
```diff
@@ -1764,6 +1761,7 @@ void PoeditFrame::OnUpdateFromPOTUpdate(wxUpdateUIEvent& event)
 #ifdef HAVE_HTTP_CLIENT
 void PoeditFrame::OnUpdateFromCrowdin(wxCommandEvent&)
 {
+    m_syncing = true;
     DoIfCanDiscardCurrentDoc([=]{
         CrowdinSyncFile(this, m_catalog, [=](std::shared_ptr<Catalog> cat){
             m_catalog = cat;
```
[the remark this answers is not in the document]
tetatetit: 508e4f920401ff39466007adf66754b730118123
- Fixed
- Improved/refactored
  - Implementation
  - Comment

Sorry for bug and unclear description and implementation creating so much confusion yet. 
Now it should be much more clear, at least at it seems to me or otherwise please feel free to ask any more questions.

src/crowdin_client.cpp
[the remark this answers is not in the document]
tetatetit: But still c65cfeff45c01ad311f50c9642fe23463c2e1042

src/crowdin_client.cpp
```diff
@@ -196,20 +196,15 @@ dispatch::future<CrowdinClient::UserInfo> CrowdinClient::GetUserInfo()
...
-                name = d["lastName"];
-                if(name.is_string())
-                    u.name += str::to_wstring(name);
+                std::string firstName, lastName;
+                try { firstName = d.at("firstName"); } catch(...) {}
+                try { lastName = d.at("lastName"); } catch(...) {}
+                u.name = str::to_wstring(firstName + ' ' + lastName);
             }
-            if(u.name.empty())
+            if(u.name.empty() || u.name == L" ")
```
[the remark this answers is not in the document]
tetatetit: Refactored and commented d81778a21f556f1e0eaa2691f3fb65fe00d3c446 what hopefully will make it more clear an self-explanatory for you and anybody else.

> How does this "fix GetUserInfo()" and why is it done?

Because earlier implementation might crash if e.g. `firstName` or `lastName` are either `null` or  missing (what is the one of above cases of `First` or `Last` name field is set/left empty in Crowdin Enterprise account settings. And above happened for me at least once.

> value("key", "default") is appropriate here

It raises exception when there is e.g. `"key": null` in JSON response what happened for me too as `"firstName": null"` or `"fullName": null` - already don't remember which exactly but due of such sometime undefined/unforeseen response format (might be due v2 API are still in Beta or so) - I decided to solve it with `try-catch` drastically :) (i.e. to make `GetUserInfo()` fully tolerant to **any** surprises with JSON response format/structure already)

> silently eating catch-all less so

But what error/message or even log entry are you expect here if most of catches just mean that field is empty or so but not actual error? While to figure out whether it's some unexpected response from server or something in our handling logic - is complicated here sometime or at least not worthing efforts since does not add any value for customer.  This again due of not absolutely clearly defined above mentioned variations of response from the server - i.e. when some of above keys are missing and when `null` what still potentially might change during v2 API beta I guess. Though if you prefere still some log entry there (just like "Missing First name") - please let me know so I will add them.

src/edframe.cpp
```diff
@@ -962,7 +962,7 @@ bool PoeditFrame::NeedsToAskIfCanDiscardCurrentDoc() const
 }
 
 template<typename TFunctor>
-void PoeditFrame::DoIfCanDiscardCurrentDoc(TFunctor completionHandler)
+void PoeditFrame::DoIfCanDiscardCurrentDoc(TFunctor completionHandler, bool bSkipUpload)
```
vslavik, line 965: Sorry, but this is an obvious hack working *around* whatever the issue is (this is unclear to me without repro steps) instead of *fixing* it where it was caused.

Can you please provide steps to reproduce so that I can better understand what is this fixing?

tetatetit: Did you see [this my comment/answer and commit mentioned there](https://github.com/vslavik/poedit/pull/633#discussion_r439735130)?  And paricularly [this comment in the code ](https://github.com/vslavik/poedit/commit/508e4f920401ff39466007adf66754b730118123#diff-52c8dd65648ccef63c74b9aa24e24da6R1771). Sorry if it's still just not enough clear/obvious to reproduce by so more detailed step-by-step is:
1. In the released `v2.3` Poedit open any `PO` file from Crowdin (via `Open from Crowdin` or just downloaded in browser via Crowdin UI)
2. Make any changes to any translation in that file in Poedit
3. Press `Sync`
4. Press `Yes``Save file before sync`
    - Result: If you put breakpoint or log entry in `CrowdinClient::UploadFile()` - you will notice it's called twice on this step (i.e. after above `OK`/`Yes`

src/http_client_casablanca.cpp
```diff
@@ -182,7 +182,6 @@ class http_client::impl
     dispatch::future<::json> get(const std::string& url)
     {
         http::http_request req(http::methods::GET);
-        req.headers().add(http::header_names::accept,     L"application/json");
```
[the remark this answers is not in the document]
tetatetit: Please see my added comment in the code a37bb61917c5659fc736ea68244499c6b9f6cab8 for it. I already commented it earlier but then moved comment into commit message while later squashed that commit with some other related bigger. Sorry losing that comment and causing confusion for you

src/http_client_casablanca.cpp
```diff
@@ -217,19 +217,29 @@ class http_client::impl
...
+            wxLogTrace("poedit.http_client", "Downloading file: closing");
             return fileStream->close();
         });
     }
 
-    dispatch::future<::json> post(const std::string& url, const http_body_data& data)
+    dispatch::future<::json> post(const std::string& url, const http_body_data& data, const http_client::headers& hdrs)
     {
         http::http_request req(http::methods::POST);
-        req.headers().add(http::header_names::accept,     L"application/json");
```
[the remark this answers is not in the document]
tetatetit: Same as https://github.com/vslavik/poedit/pull/633#discussion_r439773442

src/gexecute.cpp
```diff
@@ -23,6 +23,7 @@
  *
  */
 
+#include <wx/defs.h>
```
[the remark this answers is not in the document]
tetatetit: Well, I just tried if below UI bugs related to Crowdin functionalities would go away with [one latest releases of wxWidgets](https://github.com/wxWidgets/wxWidgets/releases/tag/v3.0.5.1) and Poedit project failed to build without above `#include`, at least on Linux, while built well with `wx/deps`. But did not solve any of below issues

In dropdown lists of `Open from Crowdin dialog (on Linux only, while no problems on other platforms with _exactly same projects_):
- problems with displaying long file paths
- lot of blank lines at the top of list in some cases if there are lot of files in dropdown
- name of picked file appears empty on some projects seems where there are lot of files at all and many of them are with long paths

Also unrelated to above dropdown:
- Very small crowdin logo in dialogs where it displayed

I will try to find time to report all above issues to [Issues](https://github.com/vslavik/poedit/issues) of this repo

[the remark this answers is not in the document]
tetatetit: Anyway please feel free to drop this commit for a while when will be merging. Or let me know if I just need to change it description or to drop it on my own and force-push this PR branch (what I would do already after resolving other issues to do not break links between already existing comments/conversations and actual code in GitHub they are made for)

src/crowdin_gui.h
tetatetit: > I don't understand why this is done. It is committed as part of "Crashed on open cached Crowdin file if logged out", but none of it seems to be related to crashing anything, and surely crashes can be avoided in a simpler way

The same answer as https://github.com/vslavik/poedit/pull/633#discussion_r433356206 . So yep, it's definitely possible to solve it as simple as variant 2 in my above comment if you will decide finally it's most proper solution (i.e. just to silently save locally without bringing `Sign in to Crowdin` UI)

src/cloud_sync.h
tetatetit: Forgot to specify steps to reproduce the crash:
1. Open any file from Crowdin in Poedit
2. Sign out from Crowdin without exiting (closing) Poedit (or probable might crash even if close and then re-open same file from local cache but don't remember exactly)
3. `Save` file (by `Ctrl+S` or `Save` in toolbar, or `File` \ `Save` - the way does not matter)

src/crowdin_client.cpp
tetatetit: Addressed 9bc52ba2523397184c28c5f4184a00908ab103ae though noticed few other `mb_str()` in _your_ code

vslavik: > Addressed 9bc52ba though noticed few other mb_str() in your code .

I should have been more precise, sorry. The issue is no `mb_str()` as such, but using it with the default value for its argument, i.e. `wxConvLibc`, which performs the conversion using whatever the standard library sees as appropriate, which is related to the current locale. When we know the charset (here, ASCII, i.e. subset of UTF-8), it's best to be explicit. When possible, it's best to avoid relying on runtime-dependent charset.

I went to fix the other code right away (I truly appreciate such remarks!) , but  I see 4 uses of `mb_str` in the current tree. 3 of them use explicit conversion object (of which 1 deals with incorrectly encoded data). 1 is the bad implicit use, but it is when passing data to a function with const char* API, `u_setDataDirectory`, where it's the only option - that library is using libc locale.

src/crowdin_client.cpp
tetatetit: Addressed cc589293a710f0f0ecfaed767de94f3fe829edd0 . Sorry that have not implemented it initially, just hurried up to get features themself working and testable.

src/crowdin_gui.cpp
tetatetit: Yep, since we need to know remote (Crowdin side) format/extension of locally opened `Xliff` file to provide correct extension here https://github.com/tetatetit/poedit/blob/cc589293a710f0f0ecfaed767de94f3fe829edd0/src/crowdin_client.cpp#L400 otherwise error response received. And at least me was not able to find a better solution than just to add `.xliff` extension to remote filename with included it remote extension as well when saving locally. Of course I considered just _replacing_ remote extension with `.xliff` but how remote e.g. `some_file_with_same_base_name` and `some_file_with_same_base_name.xliff` would be stored locally in that case without collision while both would become  `some_file_with_same_base_name.xliff` with such approach? And how to distinguish between their remote extensions/formats if opened locally and then uploaded (while correct extension should be provided as mentioned above). That is why it happens that remote `some_file_with_same_base_name.xliff` will be saved as `some_file_with_same_base_name.xliff.xliff` locally while `some_file_with_same_base_name` as `some_file_with_same_base_name.xliff`. I admit it's kind of hack but if you have any other ideas - I would highly appreciate for sharing them here. Many thanks.

P. S. For `.po` files everything is much simple - since remote `.po` is always local `.po`

src/crowdin_gui.cpp
tetatetit: Addressed 085f4795857fc4df456ecf06c3f7da23d83c11bc . Sorry for just missing to remove initially

src/crowdin_gui.h
tetatetit: > There's also a commit that seems related with a message "Move AttachCloudSync() to GUI code" which does no such thing at all and which perhaps seems be describing this commit, but in that case again, I don't understand why is it needed.

The purpose of this commit is to check (for necessity) and attach (if necessary) cloud sync in each `WriteCatalog()` call doing it in unified way regardless if it's `.xliff`, `.po` and whether opened from local cache or via `Open from Crowdin`. Just forgotten to bring it title/description appropriately when just rebased/refactored/rearranged some another commits where initially cloud sync was attached from some non-GUI code (`SetFileName()` of some of `XLIFFCatalog` classes or so) while those another commits (where it was placed somewhere in non-GUI) successfully gone during that process.  Since earlier it certainly did not work for `Xliff` (since it was not expected/supposed to sync/upload) while cloud sync was attached the only from here  https://github.com/vslavik/poedit/blob/7a0e93fff167ae5aa786d7b74f7e16b78464c88c/src/edframe.cpp#L1102 , i.e. the only from `Open from Crowdin` (so I just don't recall whether it worked for the case when `.po` opened from local cache or especially when downloaded from Crowdin UI in browser and if still worked - then how).

tetatetit: Though if you have idea for better place than it is currently is - I would highly appreciate for sharing it here.

src/edframe.cpp
```diff
@@ -1098,8 +1097,6 @@ void PoeditFrame::OnOpenFromCrowdin(wxCommandEvent&)
     DoIfCanDiscardCurrentDoc([=]{
         CrowdinOpenFile(this, [=](wxString name){
             DoOpenFile(name);
-            if (m_catalog)
-                m_catalog->AttachCloudSync(std::make_shared<CrowdinSyncDestination>());
```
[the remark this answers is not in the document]
tetatetit: Same answer/explanation as here https://github.com/vslavik/poedit/pull/633#discussion_r440335790 . Now it just well thinked out before moving, tested and works perfectly in all expected cases (`.xliff` when opened from Crowdin and from local cache, `.po` when opened from Crowdin, local cache or downloaded from Crowdin UI in browser). Though if you still see any potential problem with it now - please `Unresolve` this issue

src/crowdin_gui.cpp
tetatetit: Addressed 69df2aa7dd6996e4505240e263d0f1dac167b489 , including more clear comment now (though please feel free to `Unresolve` if still unclear)

src/crowdin_gui.cpp
tetatetit: Addressed in 3792a79dc0ceae1329e70a2790f14e3c4ad3653f by making above error messages really clear, informative, precise and fully explanatory what IMO still might be really useful for customers to understand what happened, what the problem is and how to solve it (without looking into logs) while they does not load minds of customers too much by any technical details or so. Though please feel free to `Unresolve` this conversation if you are still disagree.

src/crowdin_gui.cpp
tetatetit: Just if you @vslavik reviewed and approved it - please look once more since I have found and removed some more unsupported stuff related to the same peace of code in this commit

src/catalog.cpp
```diff
@@ -597,6 +597,9 @@ void Catalog::SetFileName(const wxString& fn)
 {
     wxFileName f(fn);
     f.Normalize(wxPATH_NORM_DOTS | wxPATH_NORM_ABSOLUTE);
+    auto name = f.GetName();
+    name.ToLong(&m_crowdinProjectId);
+    name.AfterFirst('_').ToLong(&m_crowdinFileId);
```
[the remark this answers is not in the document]
tetatetit: Addressed in a711235a66f13ccf01d2f70967ec1a4b9522e203 both fragile parsing and put together all Crowdin related logic/behaviour in `Catalog` class to single `IsFromCrowdin()` while still remained `m_crowdinFileId` and `m_crowdinProjectId` since they are involved in `Save/Sync` `.po` files downloaded from Crowdin UI in browser - just to store file and project IDs appropriately on first `Save/Sync` and to do not need to retrieve  them (making few expensive Crowdin API calls) on each further `Save/Sync` (until file is closed of course). I hope that's what you expected or please feel free to `Unresolve` this conversation otherwsie

src/catalog_xliff.h
```diff
@@ -101,9 +102,6 @@ class XLIFFCatalog : public Catalog
 
     ValidationResults Validate(bool wasJustLoaded) override;
 
-    Language GetLanguage() const override { return m_language; }
-    void SetLanguage(Language lang) override { m_language = lang; }
-
     // FIXME: PO specific
```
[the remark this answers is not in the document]
tetatetit: Addressed 0dcedfad2f58e975f240a73e25e12f335d99b23d and as side effect :) recalled why I changed what you mentioned in https://github.com/vslavik/poedit/pull/633#discussion_r433356649 - it's just to get rid redundant (as it seemed to me) stuff which keeps language in separate https://github.com/vslavik/poedit/blob/7a0e93fff167ae5aa786d7b74f7e16b78464c88c/src/catalog_xliff.h#L121 while there is https://github.com/vslavik/poedit/blob/7a0e93fff167ae5aa786d7b74f7e16b78464c88c/src/catalog.h#L384 in the base class already what confused at least me . So now I recalled there are 99% that no side effects or bugs/problem back when reverted back with this commit. So I am leaving completely up to you to decide whether to keep this commit so it it will revert back that `XLIFFCatalog::m_language` related stuff or to drop it if you still find sense in my intention/explanation/motivation

src/catalog_xliff.cpp
tetatetit: Still addressed https://github.com/vslavik/poedit/pull/633#discussion_r442268381

src/edframe.cpp
vslavik: I created an issue for this (bSkipUpload) as there seems to be more going on (race, bad UI) — see #638

src/catalog.cpp
```diff
@@ -602,8 +602,10 @@ void Catalog::SetFileName(const wxString& fn)
 
 bool Catalog::IsFromCrowdin() const
 {
-    if((m_crowdinFileId > 0 && m_crowdinProjectId) > 0
```
tetatetit: Yep, that was definitely my typo during refactoring and I would never make above intentently :), it intended to be `(m_crowdinFileId > 0 && m_crowdinProjectId > 0)`

vslavik:
Thank you! I have now pushed most of the PR to the `crowdin-v2` branch; please use that as the basis for any changes. Some missing bits are because I think I can come up with better solution:

* no bSkipUpload
* no AttachCloudSync changes

I also improved formatting, fixed broken uses of gettext's `_()`, restored link attribution code to minimize changes, made some more trivial tweaks. For reference, this is a diff from what I committed to this PR, except for trivial formatting changes. In other words, applying this patch to `crowdin-v2` branch restores your PR as-is:
[from-committed-to-pr-not-applied-bits.patch.gz](https://github.com/vslavik/poedit/files/4824666/from-committed-to-pr-not-applied-bits.patch.gz)

vslavik:
I forgot to say: I also rebased on the new (macOS) networking implementation.

There's one known problem: syncing PO files fails. 

1. Click on *Collaborate on a Translation with Others*
2. Choose "Poedit Testing Sandbox" project (I invited you), language e.g. Finnish, file default.po or any of the POT files, click OK.
3. Immediately click the *Sync* toolbar button.
4. Oops:
Screen Shot 2020-06-24 at 12 55 56

I thought my changes w.r.t PO vs POT handling caused it, but it's happening with PO too.

vslavik:
> There's one known problem: syncing PO files fails.

This is fixed in 54b86a9d.

I am now seeing `X-Crowdin-Project-ID` and `X-Crowdin-File-ID` headers in PO files downloaded from Crowdin (both via API and manually). It seems that big part of 9a58ee3271fef08845e528c00daa92e99af640b3, namely `CrowdinUpdateCatalogIDsFromHeaders`, can be safely removed after all?

tetatetit:
Hi Václav,

Fixed in this new PR https://github.com/vslavik/poedit/pull/641 (using
`crowdin-v2` as base) as well as few other issues/bugs found during testing
on all platforms (see commits there for more details). One of issues
overlaps with that you already fixed as well on your own so far I see but
still improved variable scopes/captures to limit them to only where they
are really used. Please feel free to review/comment directly in PR or here
in email. Or to fix/correct/improve on your own.

And sorry for some delay due I was in the trip.

Best regards,
Serhiy

tetatetit:
>
> I am now seeing X-Crowdin-Project-ID and X-Crowdin-File-ID headers in PO
> files downloaded from Crowdin (both via API and manually). It seems that
> big part of 9a58ee3
> <https://github.com/vslavik/poedit/commit/9a58ee3271fef08845e528c00daa92e99af640b3>,
> namely CrowdinUpdateCatalogIDsFromHeaders, can be safely removed after
> all?


Yep, I am starting changes to that same PR
https://github.com/vslavik/poedit/pull/641 to engage above to eliminate
that crappy CrowdinUpdateCatalogIDsFromHeaders so you'll see once I will
commit and also will let you know by email

Also now negotiating adding of `projId` attribute somewhere to top level
tag of XLIFF files as well so the same would work for XLIFF as well (i.e.
sync/upload of XLIFF files downloaded in browser as well) so if will be
added quickly as well from Crowdin side - it would make sense to include it
in your upcoming release as well.

tetatetit:
- Done using of X-Crowdin-Project-ID and X-Crowdin-File-ID in
   https://github.com/vslavik/poedit/pull/641/commits/43870fc19a20aebe171e90d4e5c0dcf8aa8c94b8
   - Crowdin server/API team answered that probable will implement
   providing project ID for XLIFF files soon as well so probable it would be
   good to include same approach for XLIFF as well

Issue #638, opened 2020-06-21
Race condition on Crowdin upload, messes future syncing


vslavik:
As reported by @tetatetit: under following circumstances double-upload happens:

> 1. In the released v2.3 Poedit open any PO file from Crowdin (via Open from Crowdin or just > downloaded in browser via Crowdin UI)
> 2. Make any changes to any translation in that file in Poedit
> 3. Press Sync
> 4. Press Save in Do you want to save changes question dialog box
>
> Result: If you put breakpoint or log entry in CrowdinClient::UploadFile() - you will notice it's called twice on this step (i.e. after above Save pressed)

It's not quite that, in my testing:
1. couldn't be always reproduced, double upload happened sometimes, not always
2. required in-app open from Crowdin, didn't happen with files opened normally but having Crowdin metadata

Point 2 makes sense: sync on Save is done only when opening directly, and appears related.

Strangely, doing these steps repeatedly did *not* cause it again — and using Save second and more times didn't trigger syncing as it is supposed to.

This is arguably also an issue about confusing UI: there's a Sync button and Save, but under some circumstances, they do the same thing, or nearly the same thing.

vslavik:
Part of #633 proposed fix like this (note that part of that patch is probably responsible for introducing the difference in behavior in point 2 above):

```diff
diff --git a/src/edframe.cpp b/src/edframe.cpp
index 0291d25817..240aaa9668 100644
--- a/src/edframe.cpp
+++ b/src/edframe.cpp
@@ -1039,7 +1039,7 @@ bool PoeditFrame::NeedsToAskIfCanDiscardCurrentDoc() const
 }
 
 template<typename TFunctor>
-void PoeditFrame::DoIfCanDiscardCurrentDoc(TFunctor completionHandler)
+void PoeditFrame::DoIfCanDiscardCurrentDoc(TFunctor completionHandler, bool bSkipUpload)
 {
     if ( !NeedsToAskIfCanDiscardCurrentDoc() )
     {
@@ -1049,7 +1049,7 @@ void PoeditFrame::DoIfCanDiscardCurrentDoc(TFunctor completionHandler)
 
     wxWindowPtr<wxMessageDialog> dlg = CreateAskAboutSavingDialog();
 
-    dlg->ShowWindowModalThenDo([this,dlg,completionHandler](int retval) {
+    dlg->ShowWindowModalThenDo([this,dlg,completionHandler,bSkipUpload](int retval) {
[...]

PR #641, opened 2020-07-01
Crowdin v2 fixes

tetatetit:
Fixed some problems/bugs found during testing on all platforms as well as reported by @vslavik by email

### DIFF
--- a/src/catalog.cpp
+++ b/src/catalog.cpp
@@ -605,26 +605,26 @@ void Catalog::SetFileName(const wxString& fn)
     m_fileName = f.GetFullPath();
 }
 
-bool Catalog::IsFromCrowdin() const
+bool Catalog::IsFromCrowdin(int* pProjId, int* pFileId) const
 {
-    if (m_crowdinFileId > 0 && m_crowdinProjectId > 0)
-        return true;
-
-    if (m_header.HasHeader("X-Crowdin-Project") && m_header.HasHeader("X-Crowdin-File"))
-        return true;
-
-    auto name = wxFileName(m_fileName).GetName();
-    if (name.StartsWith("CrowdinSync_"))
+    if (!(m_crowdinFileId > 0 && m_crowdinProjectId > 0))
     {
-        name = name.AfterFirst('_');
-        long project_id = 0, file_id = 0;
-        if (name.ToLong(&project_id))
-            m_crowdinProjectId = (int)project_id;
-        if (name.AfterFirst('_').ToLong(&file_id))
-            m_crowdinFileId = (int)file_id;
+        auto name = wxFileName(m_fileName).GetName();
+        if (name.StartsWith("CrowdinSync_"))
+        {
+            name = name.AfterFirst('_');
+            name.ToLong(&m_crowdinProjectId);
+            name.AfterFirst('_').ToLong(&m_crowdinFileId);
+        }
     }
 
-    return m_crowdinFileId > 0 && m_crowdinProjectId > 0;
+    if (pFileId)
+        *pFileId = (int)m_crowdinFileId;
+    if (pProjId)
+        *pProjId = (int)m_crowdinProjectId;
+
+    return (m_crowdinFileId > 0 && m_crowdinProjectId > 0)
+        || (m_header.HasHeader("X-Crowdin-Project") && m_header.HasHeader("X-Crowdin-File"));
 }
 
 namespace
--- a/src/catalog.cpp
+++ b/src/catalog.cpp
@@ -616,6 +616,12 @@ bool Catalog::IsFromCrowdin(int* pProjId, int* pFileId) const
             name.ToLong(&m_crowdinProjectId);
             name.AfterFirst('_').ToLong(&m_crowdinFileId);
         }
+
+        if (!(m_crowdinFileId > 0 && m_crowdinProjectId > 0))
+        {
+            m_header.GetHeader("X-Crowdin-Project-ID").ToLong(&m_crowdinProjectId);
+            m_header.GetHeader("X-Crowdin-File-ID").ToLong(&m_crowdinFileId);
+        }
     }
 
     if (pFileId)
@@ -623,8 +629,7 @@ bool Catalog::IsFromCrowdin(int* pProjId, int* pFileId) const
     if (pProjId)
         *pProjId = (int)m_crowdinProjectId;
 
-    return (m_crowdinFileId > 0 && m_crowdinProjectId > 0)
-        || (m_header.HasHeader("X-Crowdin-Project") && m_header.HasHeader("X-Crowdin-File"));
+    return m_crowdinFileId > 0 && m_crowdinProjectId > 0;
 }
 
 namespace
--- a/src/catalog.h
+++ b/src/catalog.h
@@ -552,11 +552,9 @@ class Catalog
         virtual void SetLanguage(Language lang);
 
         /// Is the PO file from Crowdin, i.e. sync-able?
-        bool IsFromCrowdin() const;
+        bool IsFromCrowdin(int* pProjId = nullptr, int* pFileId = nullptr) const;
         // FIXME: This shouldn't be here
-        int GetCrowdinFileId() const { return m_crowdinFileId; }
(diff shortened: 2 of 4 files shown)


vslavik:
>  Fix getting proj and file IDs from filename

Sorry for having introduced that; it is fixed as a side-effect of c53547e6.

vslavik:
> Fix upload to Crowdin on Save when opened locally  … d7f17d7

See #638 — I'm having doubts about whether Poedit *should* sync on save.

vslavik:
> Fix crash on saving PO downloaded in browser 

Finally this part is outdated based on https://github.com/vslavik/poedit/pull/633#issuecomment-652560673, as we can just remove that code, no?

tetatetit:
> > Fix upload to Crowdin on Save when opened locally  … [d7f17d7](https://github.com/vslavik/poedit/commit/d7f17d73cffed9a916794c394b0e403bb54eb4bc)
> 
> See #638 — I'm having doubts about whether Poedit _should_ sync on save.

It worked same way in v2.3 Poedit and IMO it's properly/useful (why not? what disadvantages it might cause for users?)

tetatetit:
Done using of `X-Crowdin-Project-ID` and `X-Crowdin-File-ID` in 43870fc19a20aebe171e90d4e5c0dcf8aa8c94b8

tetatetit:
> > Fix crash on saving PO downloaded in browser
> 
> Finally this part is outdated based on [#633 (comment)](https://github.com/vslavik/poedit/pull/633#issuecomment-652560673), as we can just remove that code, no?

Yep, pls feel free to remove that outdated commit from this branch/PR or let me know so I would remote it on my own

vslavik:
> It worked same way in v2.3 

I don't think that's true - [this was the only AttachCloudSync use](https://github.com/vslavik/poedit/blob/7a0e93fff167ae5aa786d7b74f7e16b78464c88c/src/edframe.cpp#L1102)

> IMO it's properly/useful (why not? what disadvantages it might cause for users?)

- the linked bug you discovered
- failure to sync presents as failure to save
- confusing UI with two buttons doing the same (see the bug)
- inconsistent behavior between opening-from-crowdin and opening-from-recent-files

vslavik:
> Use `X-Crowdin-Project-ID` and `X-Crowdin-File-ID` 

Non-conflicting commit as 79e1f920 — thanks!

tetatetit:
>
> It worked same way in v2.3
>

Yep I checked, it really does not upload if to `Save` files opened from
local cache (from recent files) bu does if opened via `Open from Crowdin`
or when downloaded in browser via Crowdin UI. Sorry for confusion, but
isn't it inconsistent behaviour for `Save` that in one case it just saves
locally while in another still uploads to Crowdin?

 the linked bug you discovered


Which one bug?

failure to sync presents as failure to save



   1. Why is the same failure ok if `Open from Crowdin` and then `Save` but
   not ok if `Save` opened from local cache or downloaded in browser via
   Crowdin UI?
   2. Why not just to first Save locally and then to upload?


 confusing UI with two buttons doing the same (see the bug)



   1. Then why `Save` of the file opened via `Open from Crowdin` still
   uploads to Crowdin while `Save` of file opened from local cache or
   downloaded in browser via Crowdin UI does not? Isn't it confusing even
   more? At least for me it is. So IMO then it should not upload on `Save` in
   none case to avoid confusions and keep same logic/behaviour for `Save` in
   all cases - i.e. just to save file locally on the disk without any upload.
   2. Which one bug?


inconsistent behavior between opening-from-crowdin and
> opening-from-recent-files
>

Inconsistent when `Save` of opened from local cache (from recent files)
uploads to Crowdin or when not? And the same for `Save` downloaded in
browser from Crowdin UI. Because  I am confused here since in 2.3 `Save` of
opened via `Open from Crowdin` saves locally and uploads to Crowdin while
`Save` of downloaded in browser from Crowdin UI does not but just saves
locally? Isn't exactly such behaviour/logic inconsistent, while IMO it
would be consistent if `Save` either would not upload in all cases or
opposite would upload in all cases?

vslavik:
> Which one bug?

#638 repeatedly referenced above.

>  Isn't it confusing even more?

Yes, as I said, the UI is confusing and I need to figure out what to do. Which is why I won't be introducing any new inconsistent behaviors until I figure it out.

The current state is no worse than 2.3. Changes here can wait until after 2.4.

tetatetit:
I got you now completely. Then since it affects user experience with
Crowdin - why not just to never upload on `Save` regardless it's opened via
`Open from Crowdin`, downloaded in browser via Crowdin UI or from local
cache (resen files)? It would be a quick simple fix while would make
behaviour of `Save` related to Crowdin consistent? While later you/we would
come with some better solution if found? I see that what you propose to
keep the same as in 2.3 is not worse than 2.3. :), but it's definitely not
good as well so making `Save` just save locally (and never upload) would
make it quite consistent IMO and would solve this problem quite quickly and
easily for a while for 2.4 I think. Unless I miss something here.

vslavik:
> but it's definitely not
good as well so making `Save` just save locally (and never upload) would
make it quite consistent IMO and would solve this problem quite quickly and
easily for a while for 2.4 I think. 

Yep. There are actually *four* cases and I failed to consider them all:

1. Open from Crowdin, edit, sync back.
2. Open from recent files as a shortcut for 1., with intent to sync to Crowdin.
3. Open a manually downloaded file, with intent to sync to Crowdin. 
4. Open a file that *happens* to have Crowdin metadata, with no desire to sync.

Case 4 is e.g. when opening Poedit's own PO files — they are downloaded from Crowdin, they report as being sync-capable. But I absolutely *do not want* my changes when playing with the files to sync. They *happen* to be sync friendly, but I didn't download them from Crowdin manually for that purpose. 

So Save behavior in these cases is:

1. Save could be said to be expected to sync, hence the motivation for having it that way (it does now)
2. It's OK to sync on save.
3. It's OK to sync on save.
4. Save **MUST NOT** sync.

And then there are 4 more cases, same as above with "...but while being offline" appended.

So I don't see any other safe and consistent way to do it in 2.4 other than not syncing on Save at all, as you suggest. What I am worried about is that it is non-obvious for previous users, or even for new ones, that Save is local-only. There should be some indicator of # of not-yet-uploaded changes, or a prompt telling to sync on closing (or saving) the file, or *something*.

tetatetit:
> So I don't see any other safe and consistent way to do it in 2.4 other than not syncing on Save at all, as you suggest. What I am worried about is that it is non-obvious for previous users, or even for new ones, that Save is local-only. There should be some indicator of # of not-yet-uploaded changes, or a prompt telling to sync on closing (or saving) the file, or something.

You definitely better then me know expectations of Poedit customers, but just sharing my own expirience with Poedit about what I  tried to guess intuitively when used Poedit and tried to figure out `Save` behaviour/logic initially

So it's just IMO the best quickest fix for 2.4. would be either
- **Never** upload on `Save`, i.e. in **none** of cases 1, 2, 3/4 while only `Sync` should be used to upload in **all** 1, 2, 3, 4 cases. I.e. `Save` behaviour would be same and consistent i all cases. I.e. just **never** upload.
- Upload on `Save` in 1, 2, but **not 3/4**, while only `Sync` should be used in case of 3/4 to upload. Which should look pretty logical and intuitive as well since when customer expect file to be uploaded back when opened via `Open from Crowdin` - IMO it's logically and intuitively to expect it to be uploaded back in case it opened from local cache (`Recent`) (since customer aware that it has been opened via `Open from Crowdin` before as well).

vslavik:
> You definitely better then me know expectations of Poedit customers

Expectations are best learned and clarified through conversations like this, which I deeply appreciate...

> Upload on Save in 1, 2, but not 3/4, while only Sync should be used in case of 3/4 to upload.

I think you're right! That indeed seems to be where we should draw the line. I'll make a separate PR for this (together with addressing #638 as this is all related) and ask you to look it over if you don't mind.

vslavik:
I think this one is outdated now, right?

tetatetit:
> I think this one is outdated now, right?

Yep, if you mean this branch/PR

PR #642, opened 2020-07-04
Crowdin v2 saving consistency

vslavik:
Addresses #638, #641 discussions:

1. Files fetched from Crowdin in-app, or fetched previously and now opened from recent files, are synced when saved
2. The explicit Sync button always saves first, avoiding double-upload and any funny asynchronous business

### DIFF
--- a/src/catalog.cpp
+++ b/src/catalog.cpp
@@ -53,7 +53,6 @@
 
 #include <set>
 #include <algorithm>
-#include <regex>
 
 
 // ----------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


vslavik:
@tetatetit What do you think?

tetatetit:
> @tetatetit What do you think?

I think it's just perfect from what we can do quickly :). Thanks a lot @vslavik . Though still will be tested thoroughly to make sure it works with all formats (PO/POT/XLIFF) consistently as described above while from first glance to commits it looks like it should :).

PR #644, opened 2020-07-08
Sync XLIFF downloaded from Crowdin UI in browser

tetatetit:
By project and file ID stored in XLIFF attributes.
Just like for PO it implemented earlier

Since Crowdin team confirmed that already added `project-id` attribute to `xliff` xml node thought they need few days to make it available/released in their app/service publically/production. But I've decided to implement this useful feature in advance to get it chance to be included in v2.4 Poedit release. So I've tested to confirm it work in Poedit right now by just adding `project-id="12334"` (where `1234` replaced with real project numeric ID) to `xliff` attribute manually

### DIFF
--- a/src/crowdin_gui.cpp
+++ b/src/crowdin_gui.cpp
@@ -677,25 +677,16 @@ bool ExtractCrowdinMetadata(CatalogPtr cat,
                 : cat->GetLanguage();
     }
 
-    if (const auto xliff = dynamic_cast<XLIFFCatalog*>(cat.get()))
+    const auto xliff = std::dynamic_pointer_cast<XLIFFCatalog>(cat);
+    if (xliff)
     {
-        auto xml = xliff->GetXMLRoot();
-        int projId = xml.attribute("project-id").as_int(-1);
-        if (projId > 0)
+        if (std::strcmp(xliff->GetXPathValue("file/header/tool//@tool-id"), "crowdin") == 0)
         {
-            auto files = xml.children("file");
-            if (files.begin() != files.end())
-            {
-                int _fileId = files.begin()->attribute("id").as_int(-1);
-                if (_fileId > 0)
-                {
-                    if (projectId)
-                        *projectId = projId;
-                    if (fileId)
-                        *fileId = _fileId;
-                    return true;
-                }
-            }
+            if (projectId)
+                *projectId = std::stoi(xliff->GetXPathValue("file//@project-id"));
+            if (fileId)
+                *fileId = std::stoi(xliff->GetXPathValue("file//@id"));
+            return true;
         }
     }
     
--- a/src/crowdin_gui.cpp
+++ b/src/crowdin_gui.cpp
@@ -39,6 +39,7 @@
 #include "languagectrl.h"
 #include "str_helpers.h"
 #include "utility.h"
+#include "catalog_xliff.h"
 
 #include <wx/app.h>
 #include <wx/artprov.h>
@@ -676,6 +677,28 @@ bool ExtractCrowdinMetadata(CatalogPtr cat,
                 : cat->GetLanguage();
     }
 
+    if (const auto xliff = dynamic_cast<XLIFFCatalog*>(cat.get()))
+    {
+        auto xml = xliff->GetXMLRoot();
+        int projId = xml.attribute("project-id").as_int(-1);
+        if (projId > 0)
+        {
+            auto files = xml.children("file");
+            if (files.begin() != files.end())
+            {
+                int _fileId = files.begin()->attribute("id").as_int(-1);
+                if (_fileId > 0)
+                {
+                    if (projectId)
+                        *projectId = projId;
+                    if (fileId)
+                        *fileId = _fileId;
+                    return true;
+                }
+            }
+        }
+    }
+    
     if (hdr.HasHeader("X-Crowdin-Project-ID") && hdr.HasHeader("X-Crowdin-File-ID"))
     {
         if (projectId)
--- a/src/crowdin_gui.cpp
+++ b/src/crowdin_gui.cpp
@@ -682,10 +682,10 @@ bool ExtractCrowdinMetadata(CatalogPtr cat,
     {
         if (std::strcmp(xliff->GetXPathValue("file/header/tool//@tool-id"), "crowdin") == 0)
         {
-            if (projectId)
-                *projectId = std::stoi(xliff->GetXPathValue("file//@project-id"));
+            if (projectId) 
+                *projectId = std::atoi(xliff->GetXPathValue("file/@*[local-name()='project-id']"));
             if (fileId)
-                *fileId = std::stoi(xliff->GetXPathValue("file//@id"));
+                *fileId = std::atoi(xliff->GetXPathValue("file/@*[local-name()='id']"));
             return true;
         }
     }
--- a/src/crowdin_gui.cpp
+++ b/src/crowdin_gui.cpp
@@ -666,7 +666,7 @@ private:
 
 bool ExtractCrowdinMetadata(CatalogPtr cat,
                             Language *lang = nullptr,
-                            int *projectId = nullptr, int *fileId = nullptr, std::string* xliffRemoteFilename = nullptr)
+                            int *projectId = nullptr, int *fileId = nullptr)
 {
     auto& hdr = cat->Header();
 
@@ -688,8 +688,6 @@ bool ExtractCrowdinMetadata(CatalogPtr cat,
                     *projectId = std::stoi(xliff->GetXPathValue("file/@*[local-name()='project-id']"));
                 if (fileId)
                     *fileId = std::stoi(xliff->GetXPathValue("file/@*[local-name()='id']"));
-                if (xliffRemoteFilename)
-                    *xliffRemoteFilename = xliff->GetXPathValue("file/@*[local-name()='original']");
                 return true;
             }
             catch (...)
@@ -780,8 +778,7 @@ void CrowdinSyncFile(wxWindow *parent, std::shared_ptr<Catalog> catalog,
 
     Language crowdinLang;
     int projectId, fileId;
-    std::string xliffRemoteFilename;
-    ExtractCrowdinMetadata(catalog, &crowdinLang, &projectId, &fileId, &xliffRemoteFilename);
+    ExtractCrowdinMetadata(catalog, &crowdinLang, &projectId, &fileId);
 
     auto handle_error = [=](dispatch::exception_ptr e){
         dispatch::on_main([=]{
@@ -819,14 +816,6 @@ void CrowdinSyncFile(wxWindow *parent, std::shared_ptr<Catalog> catalog,
                 dispatch::on_main([=]{
                     dlg->Activity->Start(_(L"Downloading latest translations…"));
                 });
-
-                if (xliffRemoteFilename.empty())
-                {
-                    if (filename.GetExt().CmpNoCase("po") != 0)  // if not PO
-                        filename.SetFullName(filename.GetName());  // set remote (Crowdin side) filename extension
-                }
-                else
-                    filename = xliffRemoteFilename;
 
                 return CrowdinClient::Get().DownloadFile(
                         projectId,
--- a/src/crowdin_gui.cpp
+++ b/src/crowdin_gui.cpp
@@ -666,7 +666,7 @@ private:
 
 bool ExtractCrowdinMetadata(CatalogPtr cat,
                             Language *lang = nullptr,
-                            int *projectId = nullptr, int *fileId = nullptr)
+                            int *projectId = nullptr, int *fileId = nullptr, std::string* xliffRemoteFilename = nullptr)
 {
     auto& hdr = cat->Header();
 
@@ -680,13 +680,22 @@ bool ExtractCrowdinMetadata(CatalogPtr cat,
     const auto xliff = std::dynamic_pointer_cast<XLIFFCatalog>(cat);
     if (xliff)
     {
-        if (std::strcmp(xliff->GetXPathValue("file/header/tool//@tool-id"), "crowdin") == 0)
+        if (xliff->GetXPathValue("file/header/tool/@tool-id") == "crowdin")
         {
-            if (projectId) 
-                *projectId = std::atoi(xliff->GetXPathValue("file/@*[local-name()='project-id']"));
-            if (fileId)
-                *fileId = std::atoi(xliff->GetXPathValue("file/@*[local-name()='id']"));
-            return true;
+            try
+            {
+                if (projectId)
+                    *projectId = std::stoi(xliff->GetXPathValue("file/@*[local-name()='project-id']"));
+                if (fileId)
+                    *fileId = std::stoi(xliff->GetXPathValue("file/@*[local-name()='id']"));
+                if (xliffRemoteFilename)
+                    *xliffRemoteFilename = xliff->GetXPathValue("file/@*[local-name()='original']");
+                return true;
+            }
+            catch (...)
+            {
+                wxLogTrace("poedit.crowdin", "Missing or malformatted Crowdin project and/or file ID");
+            }
         }
     }
     
@@ -771,7 +780,8 @@ void CrowdinSyncFile(wxWindow *parent, std::shared_ptr<Catalog> catalog,
 
     Language crowdinLang;
     int projectId, fileId;
-    ExtractCrowdinMetadata(catalog, &crowdinLang, &projectId, &fileId);
+    std::string xliffRemoteFilename;
+    ExtractCrowdinMetadata(catalog, &crowdinLang, &projectId, &fileId, &xliffRemoteFilename);
 
     auto handle_error = [=](dispatch::exception_ptr e){
         dispatch::on_main([=]{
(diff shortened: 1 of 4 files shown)


src/crowdin_gui.cpp
```diff
@@ -676,6 +677,28 @@ bool ExtractCrowdinMetadata(CatalogPtr cat,
                 : cat->GetLanguage();
     }
 
+    if (const auto xliff = dynamic_cast<XLIFFCatalog*>(cat.get()))
```
vslavik, line 680: `std::dynamic_pointer_cast<>` is your friend.

Also, please don't use this sort of valid-but-cryptic constructs like assignment within `if` with reliance on implicit null check.

src/crowdin_gui.cpp
```diff
@@ -676,6 +677,28 @@ bool ExtractCrowdinMetadata(CatalogPtr cat,
...
+                int _fileId = files.begin()->attribute("id").as_int(-1);
+                if (_fileId > 0)
+                {
+                    if (projectId)
+                        *projectId = projId;
+                    if (fileId)
+                        *fileId = _fileId;
+                    return true;
+                }
+            }
```
vslavik, line 698: This part, being relatively expensive, should be gated by `if (fileId)`

src/crowdin_gui.cpp
tetatetit: Corrected https://github.com/vslavik/poedit/pull/644/commits/87e824ed0fa0be42e13e46847bfc1119a645ad06#diff-a1865a3025926eb82afbd7f4d9f13518R680

src/crowdin_gui.cpp
tetatetit: Optimized https://github.com/vslavik/poedit/pull/644/commits/45f90046cdbc69a8773fb2c6c69ee6c3a3e9f5f2#diff-a1865a3025926eb82afbd7f4d9f13518R684 that attribures are not converted to integers in case they should not be stored to `projId` and `fileId` but IMO they _both_ need to be checked for presence in any case since without it we can not surely treat that XLIFF file as "from Crowdin" and `return true` (i.e. if at least either project or file ID is present). Unless I miss something or if you have better ideas how to optimize it but keeping same level of insurance.

tetatetit:
> Nitpicky comments below aside, I think it is fundamentally wrong idea to do low-level XML processing in Crowdin-specific code; especially without checking for XLIFF version being 1.2.

Why to check for XLIFF version if just checking for presence of `x-crowdin-project-id` and 'x-crowdin-file-id` via `XPath` should be enough how it's refactored now in 45f9004 ? Since now it can not be set by any other tool except Crowdin in any case

> In this light either [be74a59](https://github.com/vslavik/poedit/commit/be74a59f456a50b34060e43f07c376cdfd0a2303) refactoring was a mistake (without it, the natural place for this would be in `XLIFF1Catalog`) or a high-level interface in `XLIFFCatalog` is needed.

IMO virtual `IsFromCrowdin(int* projId = nullptr, int* fileId = nullptr)` in `Catalog` would be most appropriate and future proof decision which would implement only [checks for project and file IDs in filename ](https://github.com/vslavik/poedit/blob/45f90046cdbc69a8773fb2c6c69ee6c3a3e9f5f2/src/crowdin_gui.cpp#L708) which is common for all formats) while overrides of above method would implement [check for headers](https://github.com/vslavik/poedit/blob/45f90046cdbc69a8773fb2c6c69ee6c3a3e9f5f2/src/crowdin_gui.cpp#L699) in `POCatalog` and [for attributes](https://github.com/vslavik/poedit/blob/45f90046cdbc69a8773fb2c6c69ee6c3a3e9f5f2/src/crowdin_gui.cpp#L680) in `XLIFFCatalog` while both would call base method `Catalog::IsFromCrowdin()` at the beginning or at the end (if not detected by headers/attributes). Since this would clearly separate, specialize and encapsulate code responsible/applicable for particular format handling instead of mixing/putting everything together for all formats. So I can even refactor this way if you agree. But it's up to you to keep that way it's refactored now if you think it's enough for now if you think you will come with better solution later.

> I think the latter is the better approach and it would make sense to use pugixml's [XPath capabilities](https://pugixml.org/docs/manual.html#xpath.select) to add e.g.
> 
> ```c++
> std::string XLIFFCatalog::GetXPathValue(const char *xpath) const;
> ```
> 
> and use it as (pseudocode)
> 
> ```c++
> auto xliff = std::dynamic_pointer_cast<XLIFFCatalog>(cat);
> if (xliff)
> {
>     if (projectId)
>         *projectId = std::stoi(xliff->GetXPathValue("@project-id"));
>     if (fileId)
>         *fileId = std::stoi(xliff->GetXPathValue("file@id"));
> }
> ```

Refactored with `XPath` but in slightly simplified way since it's not so clear for me why to add one more public function to `XLIFFCatalog` working with `pugixml` while `GetXMLRoot()` provides access to `XPath` features as well and even more (unless you just want to drop it in favour of higher level like `GetXPathValue()` or so

> In an ideal world, this could also be used to check if the file was produced by Crowdin. For example, Xcode-produced files, with this standard header:
> 
> ```
> <xliff …>
>   <file …>
>     <header>
>       <tool tool-id="com.apple.dt.xcode" tool-name="Xcode" tool-version="10.0" build-num="10A255"/>
>     </header>
>     <body>
>     …
> ```
> 
> can then easily be checked with
> 
> ```c++
> if (xliff->GetXPathValue("file/header/tool@tool-id") == "com.apple.dt.xcode") { … }
> ```

Don't see why just checking for presence of `x-crowdin-project-id` and `x-crowdin-file-id` is not enough (since now those attributes are definitely unique for Crowdin, not too general so can not be set by any other tool except Crowdin). With how it was earlier (like `project-id` and even more just `id` for file - I would fully agree. Just changing name of project and file ID attributes to above (or any other of e.g. your proposal to better mach XLIFF spec) is not finally confirmed by Crowdin team but I would advocate it and discuss with them to change (since it's just point of changing name in what already implemented) if you agree with such approach and naming convention or would propose better

vslavik:
> Why to check for XLIFF version if just checking for presence of x-crowdin-project-id

I was uneasy about manually traversing a file with a different structure externally. The `x-crowdin-` prefix (which *still* violates the XLIFF spec, but at least is close in spirit to how the spec handles extension values) is a new development.

> Since this would clearly separate, specialize and encapsulate code responsible/applicable for particular format handling instead of mixing/putting everything together for all formats. 

Such things are always somewhat subjective, but I think having `Catalog::IsFromCrowdin()`, as I did originally, is *against* the principle of encapsulation except in the narrow sense of handling access to some headers — that would be best addressed by abstracting `Headers` (currently PO-only) to access other stuff in other formats, which is an item on my TODO list.

But besides that, *higher* level encapsulation — i.e. "all things Crowdin" makes more sense to me, i.e. putting all Crowdin-related stuff together. Consider:

1. What if there were 20 sync platforms? Would it still make sense to you to have helpers for all of them in `Catalog`?
2. Changing Crowdin to v2 API in itself (sans XLIFF) is an implementation detail that ideally no other part of Poedit would notice. But because it changes file IDs, it required changes in `Catalog` for handling the new IDs format. It wasn't well-encapsulated in Crowdin* classes, it leaked out.
3. Before the refactoring to put everything in Crowdin files, an important bit was not encapsulated and was implemented in two separate places: cached filenames were constructed in crowdin_gui.cpp code and parsed in catalog.cpp. Making changes in one of those places and not other would break things; now they are in two functions nearby in the same file — encapsulated.

> Don't see why just checking for presence of x-crowdin-project-id and x-crowdin-file-id is not enough (since now those attributes are definitely unique for Crowdin, not too general so can not be set by any other tool except Crowdin)

Presumably `<tool>` (which is confirmed) will remain canonical, while metadata deeper in the file may change with v3 API or otherwise…

> Refactored with XPath but in slightly simplified way since it's not so clear for me why to add one more public function to XLIFFCatalog working with pugixml while GetXMLRoot() provides access to XPath features as well and even more (unless you just want to drop it in favour of higher level like GetXPathValue() or so

I think this really missed the point — which is to *not* expose internal representation of `XLIFFCatalog` to outside code besides the fact that it is XLIFF and thus XML. There's little point in doing manual XPath-via-pugixml instead of manual node traversal, it gains nothing. It is *hiding* the pugixml implementation detail by way of having a string-taking, string-returning `GetXPathValue` that is the point of the suggestion.

tetatetit:
> > Why to check for XLIFF version if just checking for presence of x-crowdin-project-id
> 
> I was uneasy about manually traversing a file with a different structure externally. 

Actually structure is same affecting Crowdin specific custom attributes in both XLIFF `1.2` and `2.0`  though if you still think it makes sense to add version check even after switching to XPath (which simplifies check for presence of Crowdin specific custom attributes, i.e. eliminating necessity of any complex traverse) and it worths in term of performance - please let me know so I will add it

> The `x-crowdin-` prefix (which _still_ violates the XLIFF spec, but at least is close in spirit to how the spec handles extension values) is a new development.

Yep, as discussed with Crowdin team today - they will consider my proposal to switch to fully valid/canonical XML+XLIFF spec to use namespaces (e.g. `<xliff xmlns:crowdin="crowdin">` and then `<file crowdin:project-id="11111" crowdin:id="22222">` or please correct me if it still would not ideally match any spec and you have better suggestions/ideas) but  probably not so soon to work this way right after v2.4 Poedit release. That is why they don't see sense to switch to `x-crowdin-` prefixed names of Crowdin attributes especially if it violates specs as well and considered to be switched to namespaces in the future as final and future proof. So I've changed back to [`project-id`](https://github.com/vslavik/poedit/pull/644/commits/01473f020ed5b9eff2228e510950fbe437cba4d1#diff-a1865a3025926eb82afbd7f4d9f13518R686) and [`id`](https://github.com/vslavik/poedit/pull/644/commits/01473f020ed5b9eff2228e510950fbe437cba4d1#diff-a1865a3025926eb82afbd7f4d9f13518R688) just now `project-id` is in `file` node they decided it more future proof (anyway it will work when wil be switched to namespace based `crowdin:id` or `crowdin:file-id` attribute as final)

> > Don't see why just checking for presence of x-crowdin-project-id and x-crowdin-file-id is not enough (since now those attributes are definitely unique for Crowdin, not too general so can not be set by any other tool except Crowdin)
> 
> Presumably `<tool>` (which is confirmed) will remain canonical, while metadata deeper in the file may change with v3 API or otherwise…

Now it makes sense for me, thanks a lot for explanation. So added this check https://github.com/vslavik/poedit/pull/644/commits/01473f020ed5b9eff2228e510950fbe437cba4d1#diff-a1865a3025926eb82afbd7f4d9f13518R683

> > Refactored with XPath but in slightly simplified way since it's not so clear for me why to add one more public function to XLIFFCatalog working with pugixml while GetXMLRoot() provides access to XPath features as well and even more (unless you just want to drop it in favour of higher level like GetXPathValue() or so
> 
> I think this really missed the point — which is to _not_ expose internal representation of `XLIFFCatalog` to outside code besides the fact that it is XLIFF and thus XML. There's little point in doing manual XPath-via-pugixml instead of manual node traversal, it gains nothing. It is _hiding_ the pugixml implementation detail by way of having a string-taking, string-returning `GetXPathValue` that is the point of the suggestion.

Implemented that way 49c8c07 , thanks a lot for explanation, now it makes sense for me.

tetatetit:
Also added two project and file ID retrieval improvements  d3818d4

- Future proof namespace agnostic for the case if namespace will be added
  by Crowdin to above attributes (e.g. `crowdin:project-id`,
  `crowdin:id`) to fully match spec
- Use `std::atoi()` instread of `std::stoi()` to avoid exception if any
  of above attributes is not present so empty string is returned by
  `GetXPathValue()`

src/crowdin_gui.cpp
```diff
@@ -676,6 +677,19 @@ bool ExtractCrowdinMetadata(CatalogPtr cat,
                 : cat->GetLanguage();
     }
 
+    const auto xliff = std::dynamic_pointer_cast<XLIFFCatalog>(cat);
+    if (xliff)
+    {
+        if (std::strcmp(xliff->GetXPathValue("file/header/tool//@tool-id"), "crowdin") == 0)
```
vslavik, line 683: I must be missing something, why is `tool@tool-id` not working?

I'd rather `GetXPathValue` returned `std::string` for easy comparison of the return value. If returning `const char*`, it definitely should return `nullptr` and not `""` on not found.

vslavik:
> Future proof namespace agnostic

I take it you think it's worth it, i.e. likely to happen?

tetatetit:
> > Future proof namespace agnostic
> 
> I take it you think it's worth it, i.e. likely to happen?

Yep, as of discussion with Crowdin team. It makes sense for them so they added it as idea for consideration and probable future implementation, though still not in the nearest update when `project-id` and `tool` will appear.

src/crowdin_gui.cpp
tetatetit: > I must be missing something, why is `tool@tool-id` not working?

Yep, it does not (tried) since needs full path to node as well as `/` before `@` (single `/` is enough, not double, corrected in 8767181)

> I'd rather `GetXPathValue` returned `std::string` for easy comparison of the return value. If returning `const char*`, it definitely should return `nullptr` and not `""` on not found.

Done 8767181

tetatetit:
- Also there are other major fix in 8767181 (see it message)
- While testing above - discovered that after sync (reproduced at least on files downloaded in browser via Crowdin UI) - it does not save remote changes to the disk (i.e. to the same opened and synced local file), even if to close Poedit. The only option to save in that case to press `Save` explicitly.

tetatetit:
BTW, now Crowdin already sets `project-id` and `tool` so now it tested in live, fixed dc98593 to work original/source (remote) file is native XLIFF as well

vslavik:
I don't understand the latest changes here either :-(

> Fixes/improvement  … 8767181
> - Failed to download during sync due of misleading extension
> - `GetXPathValue()` return std::string

This is a vaguely-titled ("fixes/improvements") commit that apparently does two unrelated things inside. One of them, "Failed to download during sync due of misleading extension" is a complete and utter mystery to me. When/why/how? Then a later commit, which does a *third* thing, *also* undoes part of this. So to pick the uncontroversial bits, I have to manually untangle them and re-create your commit in a text editor.

Seriously, one commit per one thing is a great invention. It makes things *easy*. If you then want to undo a part, you don't need to do it as part of unrelated commit. You just do `git rebase -i` and drop the no-longer-wanted commit. Done. Or you do a fixup commit (one, fixup only, not mixed with other stuff) and then `git rebase -i` can be used to squash the original and the fixup together for clean history. No manual untangling.

> `exportAsXliff: true` API param for native XLIFF as well  …

This part is… complicated. On one hand, it is a good *temporary* solution for the broken v2 API that downloads corrupted files. On the other hand, I'm surprised the server-side fix, which is supposed to exist, didn't make it into production yet when other stuff did, so I'm not sure if it's worth having a temporary fix in the first place.

As a *permanent* solution, though, this shouldn't be done. All the arguments for downloading files in their original form that apply to POs *also* apply to XLIFFs. Crowdin's XLIFF-converted-to-XLIFF looses information. It's an imperfect representation of the original. It is a fallback solution to editing the real thing.

For example, [this Xcode-made XLIFF file](https://crowdin.com/translate/poedittest/60/en-cs) differs in significant ways:
* `<trans-unit>` IDs are lost; they are sometimes meaningful (e.g. id="CFBundleName" source="${PRODUCT_NAME}" — important information there)
* `<note>`s are distorted, with "Context: " prepended to them
* information about Xcode being the tool is lost; Poedit doesn't yet use it, but should, to handle IDs and notes better
* information about two embedded sub-`<file>`s is lost altogether (again: Poedit doesn't *yet* use it)

tetatetit:
> I don't understand the latest changes here either :-(
> 
> > Fixes/improvement  … [8767181](https://github.com/vslavik/poedit/commit/87671817c536427798bacb2ae1104ba9b4299aba)
> > 
> > * Failed to download during sync due of misleading extension
> > * `GetXPathValue()` return std::string
> 
> This is a vaguely-titled ("fixes/improvements") commit that apparently does two unrelated things inside. One of them, "Failed to download during sync due of misleading extension" is a complete and utter mystery to me. When/why/how? Then a later commit, which does a _third_ thing, _also_ undoes part of this. So to pick the uncontroversial bits, I have to manually untangle them and re-create your commit in a text editor.

Is it now enough clear for you in #647 PR with clean/arranged commits while same diff as the current PR as of the current moment of our discussion here? If so then once all topics/comments/discussions are resolved here - please free to close this PR in favour of above. Or please feel free to let me know here if otherwise so I would still improve that above PR branch and will force-push it.

> Seriously, one commit per one thing is a great invention. It makes things _easy_. If you then want to undo a part, you don't need to do it as part of unrelated commit. You just do `git rebase -i` and drop the no-longer-wanted commit. Done. Or you do a fixup commit (one, fixup only, not mixed with other stuff) and then `git rebase -i` can be used to squash the original and the fixup together for clean history. No manual untangling.

I am fully agreeing and admitting. But just happened to due of rush in the term of weekend and wanting to do not block/delay 2.4 release. Although I see and admit that nevertheless it turns out to take even more time for you and eventually/finally for me as well (on rebasing/rearranging, unneeded discussions/explanations etc). So I deeply apologize for this and promise to continuously improve this in all my further work on Poedit

vslavik:
> But just happened to due of rush in the term of weekend and wanting to do not block/delay 2.4 release.

I figured. Unfortunately it disoriented me so much I figured it's better to get some clarity before I rush the release.

> So I deeply apologize 

Not at all, you have nothing to apologize for! 

Thanks for the updated PR, moving to →  #647 then.

tetatetit:
> >  exportAsXliff: true API param for native XLIFF as well

> This part is… complicated.
...

> for the broken v2 API that downloads corrupted files

Sorry, but I am not sure what do you mean? _Which_ exactly "v2 API" are "_broken_" and _which_ files are downloaded "_corrupted_"

> As a permanent solution, though, this shouldn't be done. All the arguments for downloading files in their original form that apply to POs also apply to XLIFFs. Crowdin's XLIFF-converted-to-XLIFF looses information. It's an imperfect representation of the original. It is a fallback solution to editing the real thing.

I am fully consent with you as it should be ideally but as of my today discussion with Crowdin team it's not going to changed in the nearest future. So as of now we have only two options as for exactly that particular case, i.e. if the original of the file downloaded **exactly in a browser** is XLIFF
1. Just to disable `Sync` for it
2. Sync it with `exportAsXliff: true` to get it syncable at least this way rather than none

I would prefer 2 unless you are certainly for 1 or propose some better solution

> For example, this Xcode-made XLIFF file differs in significant ways
...
Fully got your concern here so fully agree and admit that in case of file opened via `Open from Crowdin` initially or from local cache later we should **always** use `exportAsXliff: false` to handle it natively as is **without** any changes to it

Issue #625, opened 2020-03-11
Unable to update string(s) in Crowdin


lanurmi:
I was attempting to change the infamous (PR #624) translation of "Needs Work/work" from plural to singular for Finnish by editing Poedit's translation in Poedit 2.3.

After I change the two strings, and click "Save", all seems ok, no errors. When I click "Synchronize", still no errors, but the old translations for those two are loaded from Crowdin! The changes I just made are lost, and not visible through Crowdin's web interface either.

Apparently if I translate something that was empty initially, such translations are properly submitted to Crowdin. The issue is related to editing existing ones. I don't think this is new to 2.3, as I do believe I tried updating those very strings with an earlier version, yet they are still unchanged.

vslavik:
This is weird. I assumed there would be a pre-existing approved or upvoted existing translations, but none are, so the latest entry - i.e. what you upload now - should be effective.

That a different translation is synced back is a side-effect of how Crowdin does things, and not *necessarily* super-bad (although weird and normal-bad). I.e. submissions are not necessarily used immediately, because a preexisting translation was approved or upvoted and takes priority.

But that's not what you describe. I don't understand this behavior. Will try to reproduce. Can you tell me what did you change these translations to, so that I follow the same steps?

lanurmi:
> Can you tell me what did you change these translations to, so that I follow the same steps?

They were both "Keskeneräisiä", and I changed them to "Keskeneräinen".

vslavik:
I don't like being that guy, the lazy developer, but I was never able to reproduce the behavior myself, on multiple occasions, not even with the same string and as close to Poedit's project setup as I was able to get.

Poedit 2.4 switches over to Crowdin's new v2 API. There's a good chance that this weirdness was fixed by the migration too. *\*fingers crossed\**

PR #646, opened 2020-07-12
`Save` behaviour related to `Sync` fixes

tetatetit:
What was originally essentially refactored/improved in #642 
See commits subjects and descriptions for more details what fixed

### DIFF
--- a/src/crowdin_gui.cpp
+++ b/src/crowdin_gui.cpp
@@ -252,7 +252,7 @@ namespace
 
 inline wxString GetCrowdinCacheDir()
 {
-    return CloudSyncDestination::GetCacheDir() + "/Crowdin/";
+    return CloudSyncDestination::GetCacheDir() + wxFILE_SEP_PATH + "Crowdin" + wxFILE_SEP_PATH;
 }
 
 
--- a/src/edframe.cpp
+++ b/src/edframe.cpp
@@ -1861,12 +1861,15 @@ void PoeditFrame::OnUpdateFromCrowdin(wxCommandEvent&)
 
     CrowdinSyncFile(this, m_catalog, [=](std::shared_ptr<Catalog> cat)
     {
+        auto cloudsync = m_catalog->GetCloudSync();
         m_catalog = cat;
         EnsureAppropriateContentView();
         NotifyCatalogChanged(m_catalog);
         RefreshControls();
         WriteCatalog(GetFileName());
-        m_catalog->AttachCloudSync(std::make_shared<CrowdinSyncDestination>());
+        // Make sure to attach it _after_ above line (`WriteCatalog`)
+        // to avoid redundant upload from there (it does if cloud sync attached)
+        m_catalog->AttachCloudSync(cloudsync);
     });
 }
 
--- a/src/edframe.cpp
+++ b/src/edframe.cpp
@@ -1865,6 +1865,7 @@ void PoeditFrame::OnUpdateFromCrowdin(wxCommandEvent&)
         EnsureAppropriateContentView();
         NotifyCatalogChanged(m_catalog);
         RefreshControls();
+        WriteCatalog(GetFileName());
         m_catalog->AttachCloudSync(std::make_shared<CrowdinSyncDestination>());
     });
 }
--- a/src/edframe.cpp
+++ b/src/edframe.cpp
@@ -1865,6 +1865,7 @@ void PoeditFrame::OnUpdateFromCrowdin(wxCommandEvent&)
         EnsureAppropriateContentView();
         NotifyCatalogChanged(m_catalog);
         RefreshControls();
+        m_catalog->AttachCloudSync(std::make_shared<CrowdinSyncDestination>());
     });
 }
 


vslavik:
> Fix save remote changes to local file after `Sync`  …
> So now they are saved implicitly while before it was saved only on explcit `Save` command after `Sync`

I am not convinced this is a good idea.

(By the way, what's up with the consistent use of backtick, normally used to quote code only in Markdown, to quote *everything*?)

tetatetit:
> > Fix save remote changes to local file after `Sync`  …
> > So now they are saved implicitly while before it was saved only on explcit `Save` command after `Sync`
> 
> I am not convinced this is a good idea.

Do you prefer that after `Sync` and just close Poedit - all remote changes would be lost locally (i.e. not saved/preserved). I.e. that when customer opens that file next time (especially with no/weak internet connection) - she/he reveals there are no last remote changes which already appeared in that file after last `Sync` but now disappeared/lost?

tetatetit:
> (By the way, what's up with the consistent use of backtick, normally used to quote code only in Markdown, to quote everything?)

This is just habit, almost automatically to quote part of code or command anywhere (including exact name of command in UI as well). This also simplifies of copying pasting such text to e.g. comments on GitHub. However if it creates inconvenience for you - I  will not use particularly with you as well as probably might reconsider using of this practice with others (just if it might be not s convenient and appropriate/expected as it seems at least to me)

vslavik:
> See commits subjects and descriptions for more details what fixed


I'm going to once again reiterate the desirability of [good commit messages](https://chris.beams.io/posts/git-commit/) (seriously, that article is great). There was a lot going on in these two PRs today with force pushes and edits to clean it up. Yet in here, after only two hours the commits are *already* inseparable, because there's a new one that undos parts of the previous ones.

There are really good reasons for the "one clean commit for one thing" rule, and they're the same as for why the commit message needs to explain the context.

For one thing, it makes reviewing easier: if all I know is "fix sync on Save of files opened from cache", I have no idea *what* it fixes, *why* or *how* (`/` is a valid separator on all platforms, how the heck does that change anything?!) — not without trying to run on all platforms second-guessing what you did and hoping to trigger the same scenario.

For another, if that code causes some trouble in the future, and somebody is tempted to change it, a `git blame` invocation can provide contexts for why something seemingly strange is the way it is — if the commit explains it, that is. If it says "fix stuff", not so much.

vslavik:
> Do you prefer that after Sync and just close Poedit - all remote changes would be lost locally (i.e. not saved/preserved). I.e. that when customer opens that file next time (especially with no/weak internet connection) - she/he reveals there are no last remote changes which already appeared in that file after last Sync but now disappeared/lost?

To be clear, *if* modified file is not saved (this is not entirely clear to me - the commit message explains nothing here, and while you alluded to something of the sorts in some comment I think, I can't find it anymore in either PR), that's not OK. I am just not sure if unconditionally *saving* the file is the right option. Marking the file as modified, i.e. also prompting about unsaved work, seems more reasonable to me at first glance.

 > This is just habit, almost automatically to quote part of code or command anywhere

I do understand quoting. I don't understand why you do it with *backsticks* — which are not quotes — in *plain* text (not Markdown).

tetatetit:
> For one thing, it makes reviewing easier: if all I know is "fix sync on Save of files opened from cache", I have no idea _what_ it fixes, _why_ or _how_ (`/` is a valid separator on all platforms, how the heck does that change anything?!) — not without trying to run on all platforms second-guessing what you did and hoping to trigger the same scenario.
> 
> For another, if that code causes some trouble in the future, and somebody is tempted to change it, a `git blame` invocation can provide contexts for why something seemingly strange is the way it is — if the commit explains it, that is. If it says "fix stuff", not so much.

Thanks for great explanations and hints so now I clearly see and understand your concerns here and why it takes your extra time and efforts on entering into and reviewing commits like this. And sorry about this so going to take it into account and to continuously improve in all further work on Poedit and other projects. Would you find commit description like below enough clear and explanatory for you and others?
"Windows uses `\` as path separator by default so despite mixing it with `/`  (while appending `/Crowdin/`) will work for opening files - it will not match in `ShouldSyncToCrowdinAutomatically()`" 
Or otherwise please provide your example of commit description to let me better understand what you expect here

tetatetit:
> > Do you prefer that after Sync and just close Poedit - all remote changes would be lost locally (i.e. not saved/preserved). I.e. that when customer opens that file next time (especially with no/weak internet connection) - she/he reveals there are no last remote changes which already appeared in that file after last Sync but now disappeared/lost?
> 
> To be clear, _if_ modified file is not saved (this is not entirely clear to me - the commit message explains nothing here, and while you alluded to something of the sorts in some comment I think, I can't find it anymore in either PR), that's not OK. I am just not sure if unconditionally _saving_ the file is the right option. Marking the file as modified, i.e. also prompting about unsaved work, seems more reasonable to me at first glance.

Personally me and IMO I would consider explicit `Sync` just like explicit `Save` but just more advanced. i.e. if customer already do `Sync` explicitly - then to consider this as customer readiness to get remote changes to be saved locally by default without any extra actions. Since so far I know and encountered in various application - usually `Sync` brings remote changes to local and saves them by default without any additional actions.  That is why most of customers except from `Sync`/`Synchronize` action/command in most of application. Though it's only my personal stats/observation so I might be wrong here. Just not sure how it worked in v2.3 Poedit

vslavik:
> "Windows uses \ as path separator by default so despite mixing it with / (while appending /Crowdin/) will work for opening files - it will not match in ShouldSyncToCrowdinAutomatically()"

Oh! I think there may be more problems with the overly simplistic `ShouldSyncToCrowdinAutomatically()` —  case, UNC paths, differently expressed paths etc. Let me see if there's a more "proper" way to check.

vslavik:
> Oh! I think there may be more problems with the overly simplistic ShouldSyncToCrowdinAutomatically() — case, UNC paths, differently expressed paths etc. Let me see if there's a more "proper" way to check.

Applied as-is in 5b2a7511 with added comment explaining that I implemented `ShouldSyncToCrowdinAutomatically()` badly.

vslavik:
> Personally me and IMO I would consider explicit Sync just like explicit Save but just more advanced. i.e. if customer already do Sync explicitly - then to consider this as customer's readiness to get remote changes to be saved locally by default without any extra actions.

This is a reasonable expectation, I agree. 

As counter-view, consider what if the changes synced back from Crowdin are bad, e.g. replace translator's content with another contributor's (this can happen with Crowdin, but isn't common in most sync scenarios, so that's an important difference).  

If automatic save is done, the translator has no easy way to get back to their work. If this happens *unexpectedly*, the user had no reason to make a backup before, and would crucify me for suggesting that — in their view, it would be Poedit's fault.

If saving is *not* done, then they can at least always discard unsaved changes and return to what they had before.

This isn't theoretical, by the way, I sometimes here about it. Mostly it can be explained by voting or approving at Crowdin, but is still confusing. And then there are situations like #625...

tetatetit:
> > Personally me and IMO I would consider explicit Sync just like explicit Save but just more advanced. i.e. if customer already do Sync explicitly - then to consider this as customer's readiness to get remote changes to be saved locally by default without any extra actions.
> 
> This is a reasonable expectation, I agree.
> 
> As counter-view, consider what if the changes synced back from Crowdin are bad, e.g. replace translator's content with another contributor's (this can happen with Crowdin, but isn't common in most sync scenarios, so that's an important difference).
> 
> If automatic save is done, the translator has no easy way to get back to their work. If this happens _unexpectedly_, the user had no reason to make a backup before, and would crucify me for suggesting that — in their view, it would be Poedit's fault.
> 
> If saving is _not_ done, then they can at least always discard unsaved changes and return to what they had before.
> 
> This isn't theoretical, by the way, I sometimes here about it. Mostly it can be explained by voting or approving at Crowdin, but is still confusing. And then there are situations like #625...

It also makes sense all what you mentioned above. 
1. Then how about just asking "Are you sure?" on first `Sync` of file downloaded in browser via Crowdin UI? But if you think it might provide worse UX - then I agree that just syncing remote changes to local file without saving it implicitly is less risky in the term of losing user's work (without easy way to recover thought it will still available in Crowdin's history/variants of translations).
   1. But then it should definitely mark synced file as modified and to ask for `Save` on exit, IMO (which should be actually the case for files `Open from Crowdin` as well I guess).
   2. Or here might worth to still save implicitly for files opened via `Open from Crowdin` and then from local cache.
2. If you concern about losing user's work - then how is handled the case when e.g. user edited a file and then do `Sync` without doing `Save` before it? Or is `Save` made implicitly _before_ `Sync` in such case?

vslavik:
> Then how about just asking "Are you sure?" on first Sync of file downloaded in browser via Crowdin UI?

The answer will be "yes, I am sure" — only to be regretted *after the fact* when the changed synced back would be saved irrevocably. That's the point, you don't know something bad might happen *before* doing the bidirectional sync.

> But then it should definitely mark synced file as modified and to ask for Save on exit, IMO

Yes, absolutely.

> (which should be actually the case for files Open from Crowdin as well I guess).

These are already saved as-is, so unless you touch them, there's no point in saving, i.e. shouldn't be initially marked as modified.

> If you concern about losing user's work - then how is handled the case when e.g. user edited a file and then do Sync without doing Save before it? Or is Save made implicitly before Sync in such case?

Yes. I thought that's what we agreed made sense?

tetatetit:
> > Then how about just asking "Are you sure?" on first Sync of file downloaded in browser via Crowdin UI?
> 
> The answer will be "yes, I am sure" — only to be regretted _after the fact_ when the changed synced back would be saved irrevocably. That's the point, you don't know something bad might happen _before_ doing the bidirectional sync.

Sorry for might being not obvious here. I just meant here to show message box in above case saying e.g. "WARNING: Some of your local translations in currently opened file might be overwritten with remote if edited by others. Continue?" i.e. to show this message box only on _first_ `Sync` after file is opened.

vslavik:
No, I understand. But what would you answer, as a user, to that question? You would say "yes". 

You would *still* not expect the changes to be written to the file with no possibility to undo. You would expect that at worst you won't save the result. Part of the problem is, of course, that Poedit lacks document-wide undo.

tetatetit:
> No, I understand. But what would you answer, as a user, to that question? You would say "yes".
> 
> You would _still_ not expect the changes to be written to the file with no possibility to undo. You would expect that at worst you won't save the result. Part of the problem is, of course, that Poedit lacks document-wide undo.

Ok, I see you point of view and now it makes sense. It's difficult to warn user without long complex message which might either somebody might be too rushing to read and understand completely or just would strain others.

vslavik:
We're agreed, then — I'll do it like that and merge and amend everything today.

And of course, all this can be reconsidered later if circumstances change.

vslavik:
faa35685 and 09811b93 implemented as discussed above — using marking as modified instead of unconditional save, and thus also simpler re-attachment of `CloudSyncDestination`.

Thanks!

tetatetit:
> [faa3568](https://github.com/vslavik/poedit/commit/faa35685ffa1b4de43e926716999159d92d56e21) and [09811b9](https://github.com/vslavik/poedit/commit/09811b93a6ecc024607a02094d6eaf217cff9b47) implemented as discussed above — using marking as modified instead of unconditional save, and thus also simpler re-attachment of `CloudSyncDestination`.
> 
> Thanks!

One more side effect of just marking as modified on `Sync` I missed during our discussion
1. `Sync` any file
2. `Save` explicitly or by confirming on exit
* Result: File is uploaded to Crowdin again while not really modified locally and is exactly same as remote. So no uploaded is needed in this case really so actually double upload occur in the case it really does not make sense/necessity. This could be essential problem in case of weak connection and big file, as well as increased load on Crowdin server in case it can be avoided
* Proposed fix/workaround: Add some flag to `MarkAsModified()` to distinguish whether file was modified locally or just remotely without local modification afterward so does not need to re-upload it to Crowdin on `Save`/exit.

vslavik:
> One more side effect of just marking as modified on Sync I missed during our discussion

Oh crap.

> Proposed fix/workaround:

Or go back to your original approach, i.e. just `WriteCatalog()` and be done with it, instead of fretting over hopefully-theoretical issues with that approach as I did.

I think that would be the wisest. Adding any further flag would complicate the already messy, badly designed save/sync logic even further, with even further special cases than somebody (read: me) will inevitably miss in future changes. It's walking on eggshells already as it is.

For Poedit 3, this warrants thinking over completely and, I think, getting rid of all this complicated stuff altogether: files should either be cloud-first and have just Save (which goes to cloud, with offline cache) and no Sync button; or file-first and have a Sync button and never sync automatically on save. But that's too big UI change to do in 2.4, especially at last minute.

tetatetit:
> Or go back to your original approach, i.e. just WriteCatalog() and be done with it, instead of fretting over hopefully-theoretical issues with that approach as I did.

Just to get confirmation from you and to make sure I completely understand you now. Do you mean proposed here https://github.com/vslavik/poedit/pull/646/commits/d95afb6329c890df332a7099c2714115bf5a144c#diff-52c8dd65648ccef63c74b9aa24e24da6R1870 ? If then - will you apply it on your own or can I create new PR for it?


> For Poedit 3, this warrants thinking over completely and, I think, getting rid of all this complicated stuff altogether: files should either be cloud-first and have just Save (which goes to cloud, with offline cache) and no Sync button; or file-first and have a Sync button and never sync automatically on save. But that's too big UI change to do in 2.4, especially at last minute.

IMO your initial `Sync` button is not so bad itself in the term of UX / use cases / workflows since it's not just uploads or downloads but actually does what is really usually expected from synchronization - i.e. merges remote changes with local. Though I agree that it should be re-designed / refactored / re-architected in Poedit code since I guess when you initially designed / architected / planned it - you did not expect / plan for cloud / remote sync.

PR #647, opened 2020-07-13
Sync XLIFF downloaded from Crowdin UI in browser (clean)

tetatetit:
By project and file ID stored in XLIFF attributes (just like for PO it implemented earlier)

### DIFF
--- a/src/catalog_xliff.cpp
+++ b/src/catalog_xliff.cpp
@@ -460,7 +460,7 @@ Catalog::ValidationResults XLIFFCatalog::Validate(bool)
     return res;
 }
 
-const char* XLIFFCatalog::GetXPathValue(const char* xpath) const
+std::string XLIFFCatalog::GetXPathValue(const char* xpath) const
 {
     auto x = m_doc.child("xliff").select_node(xpath);
     auto v = x.attribute().value();
--- a/src/catalog_xliff.cpp
+++ b/src/catalog_xliff.cpp
@@ -460,7 +460,17 @@ Catalog::ValidationResults XLIFFCatalog::Validate(bool)
     return res;
 }
 
-
+const char* XLIFFCatalog::GetXPathValue(const char* xpath) const
+{
+    auto x = m_doc.child("xliff").select_node(xpath);
+    auto v = x.attribute().value();
+    if (v)
+        return v;
+     v = x.node().value();
+     if (v)
+         return v;
+     return "";
+}
 
 
 
--- a/src/catalog_xliff.h
+++ b/src/catalog_xliff.h
@@ -109,6 +109,7 @@ public:
     void RemoveDeletedItems() override {}
 
     pugi::xml_node GetXMLRoot() const { return m_doc.child("xliff"); }
+    const char* GetXPathValue(const char* xpath) const;
 
 protected:
     XLIFFCatalog(const wxString& filename, pugi::xml_document&& doc)
--- a/src/catalog_xliff.h
+++ b/src/catalog_xliff.h
@@ -109,7 +109,7 @@ public:
     void RemoveDeletedItems() override {}
 
     pugi::xml_node GetXMLRoot() const { return m_doc.child("xliff"); }
-    const char* GetXPathValue(const char* xpath) const;
+    std::string GetXPathValue(const char* xpath) const;
 
 protected:
     XLIFFCatalog(const wxString& filename, pugi::xml_document&& doc)
--- a/src/crowdin_client.cpp
+++ b/src/crowdin_client.cpp
@@ -424,8 +424,8 @@ dispatch::future<void> CrowdinClient::DownloadFile(int project_id,
 
     wxString ext(file_extension);
     ext.MakeLower();
-    const bool isXLIFFNative = (ext == "xliff" || ext == "xlf");
-    const bool isXLIFFConverted = (!isXLIFFNative && ext != "po" && ext != "pot");
+
+    const bool isXLIFFConverted = (ext != "po" && ext != "pot");
 
     return m_api->post(
         "projects/" + std::to_string(project_id) + "/translations/builds/files/" + std::to_string(file_id),
@@ -445,7 +445,7 @@ dispatch::future<void> CrowdinClient::DownloadFile(int project_id,
             wxString outfile(output_file);
             file.move_to(outfile);
 
-            if (isXLIFFNative || isXLIFFConverted)
+            if (isXLIFFConverted)
                 PostprocessDownloadedXLIFF(outfile);
         });
 }
--- a/src/crowdin_gui.cpp
+++ b/src/crowdin_gui.cpp
@@ -680,12 +680,12 @@ bool ExtractCrowdinMetadata(CatalogPtr cat,
     const auto xliff = std::dynamic_pointer_cast<XLIFFCatalog>(cat);
     if (xliff)
     {
-        if (std::strcmp(xliff->GetXPathValue("file/header/tool//@tool-id"), "crowdin") == 0)
+        if (xliff->GetXPathValue("file/header/tool//@tool-id") == "crowdin")
         {
             if (projectId) 
-                *projectId = std::atoi(xliff->GetXPathValue("file/@*[local-name()='project-id']"));
+                *projectId = std::stoi(xliff->GetXPathValue("file/@*[local-name()='project-id']"));
             if (fileId)
-                *fileId = std::atoi(xliff->GetXPathValue("file/@*[local-name()='id']"));
+                *fileId = std::stoi(xliff->GetXPathValue("file/@*[local-name()='id']"));
             return true;
         }
     }
--- a/src/crowdin_gui.cpp
+++ b/src/crowdin_gui.cpp
@@ -817,9 +817,6 @@ void CrowdinSyncFile(wxWindow *parent, std::shared_ptr<Catalog> catalog,
                     dlg->Activity->Start(_(L"Downloading latest translations…"));
                 });
 
-                if (filename.GetExt().CmpNoCase("po") != 0)  // if not PO
-                    filename.SetFullName(filename.GetName());  // set remote (Crowdin side) filename extension
-
                 return CrowdinClient::Get().DownloadFile(
                         projectId,
                         crowdinLang,
--- a/src/crowdin_gui.cpp
+++ b/src/crowdin_gui.cpp
@@ -39,6 +39,7 @@
 #include "languagectrl.h"
 #include "str_helpers.h"
 #include "utility.h"
+#include "catalog_xliff.h"
 
 #include <wx/app.h>
 #include <wx/artprov.h>
@@ -676,6 +677,19 @@ bool ExtractCrowdinMetadata(CatalogPtr cat,
                 : cat->GetLanguage();
     }
 
+    const auto xliff = std::dynamic_pointer_cast<XLIFFCatalog>(cat);
+    if (xliff)
+    {
+        if (std::strcmp(xliff->GetXPathValue("file/header/tool//@tool-id"), "crowdin") == 0)
+        {
+            if (projectId) 
+                *projectId = std::atoi(xliff->GetXPathValue("file/@*[local-name()='project-id']"));
+            if (fileId)
+                *fileId = std::atoi(xliff->GetXPathValue("file/@*[local-name()='id']"));
+            return true;
+        }
+    }
+    
     if (hdr.HasHeader("X-Crowdin-Project-ID") && hdr.HasHeader("X-Crowdin-File-ID"))
     {
         if (projectId)
--- a/src/crowdin_gui.cpp
+++ b/src/crowdin_gui.cpp
@@ -682,11 +682,18 @@ bool ExtractCrowdinMetadata(CatalogPtr cat,
     {
         if (xliff->GetXPathValue("file/header/tool//@tool-id") == "crowdin")
         {
-            if (projectId) 
-                *projectId = std::stoi(xliff->GetXPathValue("file/@*[local-name()='project-id']"));
-            if (fileId)
-                *fileId = std::stoi(xliff->GetXPathValue("file/@*[local-name()='id']"));
-            return true;
+            try
+            {
+                if (projectId) 
+                    *projectId = std::stoi(xliff->GetXPathValue("file/@*[local-name()='project-id']"));
+                if (fileId)
+                    *fileId = std::stoi(xliff->GetXPathValue("file/@*[local-name()='id']"));
+                return true;
+            }
+            catch(...)
+            {
+                wxLogTrace("poedit.crowdin", "Missing or malformatted Crowdin project and/or file ID");
+            }
         }
     }
     


tetatetit:
Might be better to continue here so just duplicating https://github.com/vslavik/poedit/pull/644#issuecomment-657604181 I posed just few minutes after you already moved to here :)

> >  exportAsXliff: true API param for native XLIFF as well

> This part is… complicated.
...

> for the broken v2 API that downloads corrupted files

Sorry, but I am not sure what do you mean? _Which_ exactly "v2 API" are "_broken_" and _which_ files are downloaded "_corrupted_"

> As a permanent solution, though, this shouldn't be done. All the arguments for downloading files in their original form that apply to POs also apply to XLIFFs. Crowdin's XLIFF-converted-to-XLIFF looses information. It's an imperfect representation of the original. It is a fallback solution to editing the real thing.

I am fully consent with you as it should be ideally but as of my today discussion with Crowdin team it's not going to changed in the nearest future. So as of now we have only two options as for exactly that particular case, i.e. if the original of the file downloaded **exactly in a browser** is **exactly** XLIFF
1. Just to disable `Sync` for it
2. Sync it with `exportAsXliff: true` to get it syncable at least this way rather than none

I would prefer 2 unless you are certainly for 1 or propose some better solution

> For example, this Xcode-made XLIFF file differs in significant ways
...
Fully got your concern here so fully agree and admit that in case of file opened via `Open from Crowdin` initially or from local cache later we should **always** use `exportAsXliff: false` to handle it natively as is **without** any changes to it

tetatetit:
And BTW, while deeping inside issues attempted to be addressed in this/#644 PR - discovered issue introduced with your refactoring of my #633 in your https://github.com/vslavik/poedit/commit/ea1a6fddc1ba088b12687d212823a7f92198f6bf#diff-a1865a3025926eb82afbd7f4d9f13518R629 which could be discussed in separate PR/issue of course but so far as I see fix of it is closely related to this issue so might be the best place to discuss it here as well.
So let's start from steps to reproduce:
1. Run Poedit with `WXTRACE=poedit.crowdin` env var
2. `Open from Crowdin` some natively XLIFF file (i.e. original/source of which is XLIFF)
3. Do either `Sync` or `Save` (`Save` makes sense in any branch in which it still works before broken in #642)
4. Look into `DownloadFile(project_id` of trace output
Result: `file_extension=` there is the name of the file instead of `xliff` extension which leads to `isXLIFFConverted` is `false` which leads to `"exportAsXliff: false` what is definitely not what we want accordingly to above and https://github.com/vslavik/poedit/pull/644#issuecomment-657604181 discussion

vslavik:
> Sorry, but I am not sure what do you mean? Which exactly "v2 API" are "broken" and which files are downloaded "corrupted"

If you "Open from Crowdin" XLIFF file that is natively uploaded as XLIFF (as the example in the sandbox project), without `exportAsXliff`, it won't download the actual original, but a corrupted version with translations replaced by some weird whitespace. This is "a known bug" according to Crowdin that is "already fixed in testing". 

Please don't tell me they changed their mind on fixing it :(

vslavik:
Thanks; squashed first three commits into 983a74dc.

vslavik:
> And BTW, while deeping inside issues attempted to be addressed in this/#644 PR - discovered issue introduced with your refactoring of my #633 in your ea1a6fd#diff-a1865a3025926eb82afbd7f4d9f13518R629

Sorry about that. I'm getting the uneasy feeling that this code is crazy fragile — I as much as look at it funny and I break something :(

I don't think extracting data from XLIFF nodes is appropriate fix - it's just another patch over a patch over a patch. The real cause is, in my view, the assumption in `CrowdinSyncFile` that files *must* have the double extension; when I removed the `.xliff.xliff` double-extension, I did it unaware that some other innocuous looking code depends on it:
```c++
if (filename.GetExt().CmpNoCase("po") != 0)  // if not PO
    filename.SetFullName(filename.GetName());  // set remote (Crowdin side) filename extension
```

I think the fundamental mistake is having file extension as `DownloadFile` (and `UploadFile`) argument - this is something server-side determined, so we shouldn't rely on it. Is there an API endpoint allowing as to get metadata for a specific file given projectId+fileId?

tetatetit:
> > And BTW, while deeping inside issues attempted to be addressed in this/#644 PR - discovered issue introduced with your refactoring of my #633 in your [ea1a6fd](https://github.com/vslavik/poedit/commit/ea1a6fddc1ba088b12687d212823a7f92198f6bf)#diff-a1865a3025926eb82afbd7f4d9f13518R629
> 
> Sorry about that. I'm getting the uneasy feeling that this code is crazy fragile — I as much as look at it funny and I break something :(

Honestly speaking I got the same feeling everywhere in the Poedit code I needed to introduce Crowdin related changes except protocol (`http_client`) and API `crowdi_client` related code the only which was quite robustly abstracted from other Poedit code, Anyway it might be just not enough familiarity with your professional style/approaches/patterns/paradigms/semantics just like every engineer has his/her own (just like to one comes to somebody kitchen :) which I must to admit is much better standardized in Qt, GoLang and Rails projects in this terms). 

> I don't think extracting data from XLIFF nodes is appropriate fix - it's just another patch over a patch over a patch. The real cause is, in my view, the assumption in `CrowdinSyncFile` that files _must_ have the double extension; when I removed the `.xliff.xliff` double-extension, I did it unaware that some other innocuous looking code depends on it:
> if (filename.GetExt().CmpNoCase("po") != 0)  // if not PO
>     filename.SetFullName(filename.GetName());  // set remote (Crowdin side) filename extension
> ```
> 

It's not much about exactly above but rather about conceptual logic/approach itself (which I already mentioned to you about this "doubled extension" when you asked about on some of reviews so far I remember but I fully understand you might miss it due of so much code reviewed for which thanks a lot to you anyway) but rather about that if currently there are no well established place to keep original file extension except as in filename - then how can we distinguish between original file extension is XLIFF (`exportAsXliff: false`) and original file with no extension (`exportAsXliff: true`)? I agree that not so much extension we need to know but just whether to download with `exportAsXliff` `true` or `false` which can be encoded in the name in similar way as project and file id (which actually needed only in case of files opened via `Open from Crowdin` or from local cache later). But this just a question of preference to approach/style which I chosen just from that point to make original file extension human readable to end user (if user just look on the name of opened file). So if you don't have better ideas here - then what is wrong with doubling extension for you? :).
 
> I think the fundamental mistake is having file extension as `DownloadFile` (and `UploadFile`) argument - this is something server-side determined, so we shouldn't rely on it. Is there an API endpoint allowing as to get metadata for a specific file given projectId+fileId?

Here it is https://support.crowdin.com/api/v2/#operation/api.projects.files.post but I and Crowdin team always try to avoid any extra calls if possible to keep necessary data somewhere locally from Poedit / file name / file metadata side. Unless it adds too much complexity to the code what I don't think doubling `.xliff` extensions the case (but rather point of elegancy I guess but not a real problem/weakness/fragility on itself, at least IMO).

vslavik:
> I don't think extracting data from XLIFF nodes is appropriate fix - it's just another patch over a patch over a patch.

What I wrote there is BS. I misread it as looking for the *filename*, not using it as an indicator meaning only "this is an convert XLIFF".  You're absolutely right about that, will do. Thanks!

tetatetit:
@vslavik , just checking if I can help with anything (to answer I just missed or to do/implement) to speed up merging and closing this PR and to speed up release. e.g. https://github.com/vslavik/poedit/pull/647#issuecomment-657606313 related to https://github.com/vslavik/poedit/pull/647/commits/0795c5a6f6ffe3806ae288baa561ea845724aa22 seems unanswered/unresolved yet and if you answered with prefered solution (chosen among proposed or some another idea) - I could proceed with implementing it. Thanks for all your work and efforts!

vslavik:
In 9487af01 — extract correct filename/extension from XLIFF except that the logic slightly differs from what https://github.com/vslavik/poedit/pull/647#issuecomment-657618842 says in that if this can't be extracted, then file extension is trusted (it means file format is natively supported, therefore Poedit didn't change it).

In 4b682728 — applied the workaround for broken native XLIFF support, i.e. 0795c5a, rebased atop the above to make reverting in the future simpler; note that the `crowdin_gui.cpp` part is no longer needed thanks to 9487af01.

vslavik:
> @vslavik , just checking if I can help with anything

I'd appreciate double checking if I didn't screw up anything this time.

Thanks!

tetatetit:
> In [9487af0](https://github.com/vslavik/poedit/commit/9487af01bb1347d25c9f756f99fc0e4b7ba2aca8) — extract correct filename/extension from XLIFF except that the logic slightly differs from what [#647 (comment)](https://github.com/vslavik/poedit/pull/647#issuecomment-657618842) says in that if this can't be extracted, then file extension is trusted (it means file format is natively supported, therefore Poedit didn't change it).
> 
> In 4b682728 — applied the workaround for broken native XLIFF support, i.e. [0795c5a](https://github.com/vslavik/poedit/commit/0795c5a6f6ffe3806ae288baa561ea845724aa22), rebased atop the above to make reverting in the future simpler; note that the `crowdin_gui.cpp` part is no longer needed thanks to [9487af0](https://github.com/vslavik/poedit/commit/9487af01bb1347d25c9f756f99fc0e4b7ba2aca8).



Unfortunately it still needs to handle below special cases in case if **original/source** of downloaded file is **XLIFF**
1. As it comes from your https://github.com/vslavik/poedit/pull/644#issuecomment-657252958 starting from your "For example, [this Xcode-made XLIFF file](https://crowdin.com/translate/poedittest/60/en-cs) differs in significant ways"  and what I just tested on Crowdin to confirm - `exportAsXliff: false` should be used to download it **"as is"** just like in case with PO/POT. Since when I implemented closest logic to your in above commit for setting of `exportAsXliff` in https://github.com/vslavik/poedit/pull/644/commits/dc985930b95380cb40f410dfc0ef74fbb6aeb7d6 - I just did not expect above problem you highlighted. I.e. `&& ext != "XLIFF" && ext != "XLF"` should be added back to here https://github.com/vslavik/poedit/pull/647/commits/0795c5a6f6ffe3806ae288baa561ea845724aa22#diff-ee71506cd9832717af50c2721cc35d85L428
2. While above addition proposed by me solves above problem so XLIFF is returned really  "as is" (i.e. without any changes or missing/corruption of anything) - it causes XLIFF to do not contain `original` attribute in that case as well as any other Crowdin specific attributes/nodes (project and file id etc.) and there are nothing to do with that in the nearest time as of discussion with Crowdin team. This breaks only below use case / workflow:
   1. File downloaded from browser via Crowdin UI with "Download in XLIFF" (just recall we are talking only about the case when original/source file of downloaded file is XLIFF) which differs from simple `Download` in that `Download in XLIFF` is downloaded as already parsed and recreated by Crowdin from internal structure (i.e. XLIFF-to-XLIFF converted as you called it somewhere) while simple `Download` - downloads "as is" natively just just like in case with PO/POT with only difference that XLIFF unfortunately does not contain project and file IDs in this case as well as other Crowdin specific attributes
   2. Opened in Poedit
   3. Press `Sync`
   4. File is successfully uploaded and then downloaded back with **`exportAsXliff: false`**  (since it extension is XLIFF  - taken from `original` attribute or would be same if taken just from local filename) **but** already "as is" **without** file and project IDs attributes as well as any other Crowdin specific attributes
   * Result: `Sync` button disappears from toolbar
   * Proposed fix/workaround:  Change this https://github.com/vslavik/poedit/commit/9487af01bb1347d25c9f756f99fc0e4b7ba2aca8#diff-a1865a3025926eb82afbd7f4d9f13518L823 to somethign like
    ```
    if (!xliffRemoteFilename.empty())
    {
        auto ext = filename.GetExt();
        ext.MakeLower();
        if(ext == "xliff" || ext == "xlf)
            filename.SetExt(""); // none or any but just not XLIFF
        else
            filename.Assign(xliffRemoteFilename, wxPATH_UNIX);
    }
    ```

vslavik:
> As it comes from your #644 (comment) starting from your "For example, this Xcode-made XLIFF file differs in significant ways" and what I just tested on Crowdin to confirm - exportAsXliff: false should be used to download it "as is" just like in case with PO/POT. 

I think we're moving in circles.

Again:

1. Yes, I agree! It should behave that way. I want that.
2. Crowdin API has a bug currently. It doesn't work that way. Try it yourself: revert 4b68272, then open *Poedit Testing Sandbox →  AppSandbox.xliff* file from Crowdin. Poedit will tell you the file is empty. It shouldn't be. 
3. Hence applying 4b68272 **as a workaround for the bug** until Crowdin deploys the fix.

vslavik:
> While above addition proposed by me solves above problem so XLIFF is returned really "as is" (i.e. without any changes or missing/corruption of anything) - it causes XLIFF to do not contain original attribute in that case as well as any other Crowdin specific attributes/nodes (project and file id etc.) and there are nothing to do with that in the nearest time as of discussion with Crowdin team. This breaks only below use case / workflow:

Note that it does *not* break it as of now as 4b68272 remains applied.

> Proposed fix/workaround:

A better fix is something I in fact started doing instead of 9487af01 before I realized the file extension is still important (this also prompted my question about getting this from API as you may recall): pass `wasExportedAsXliff` as another boolean argument to `DownloadFile()`. The logic here is complicated, so the extra verboseness and clarity would only help, and it would unambiguously say what needs to be done.

tetatetit:
> > As it comes from your #644 (comment) starting from your "For example, this Xcode-made XLIFF file differs in significant ways" and what I just tested on Crowdin to confirm - exportAsXliff: false should be used to download it "as is" just like in case with PO/POT.
> 
> I think we're moving in circles.
> 
> Again:
> 
> 1. Yes, I agree! It should behave that way. I want that.
> 2. Crowdin API has a bug currently. It doesn't work that way. Try it yourself: revert [4b68272](https://github.com/vslavik/poedit/commit/4b682728c749a28b9291cce347c9dd8769caf965), then open _Poedit Testing Sandbox →  AppSandbox.xliff_ file from Crowdin. Poedit will tell you the file is empty. It shouldn't be.
> 3. Hence applying [4b68272](https://github.com/vslavik/poedit/commit/4b682728c749a28b9291cce347c9dd8769caf965) **as a workaround for the bug** until Crowdin deploys the fix.

No! It works perfectly now with my fix #648 (https://github.com/vslavik/poedit/pull/648/commits/9d68fd54d51b6017abd06bb170779d244ca2219f). See screenshots below or just try it yourself. So it looks like Crowdin fixed that problem you mentioned related to it. Of course if I did not miss something in your explanation / description / steps to reproduce or otherwise sorry if I may be just did not delve into described your problem deeper to reveal and confirm it now fixed!

image
image

vslavik:
Moving this topic over to #648. Let's please don't add any more comments on this PR unless they are about the commits applied directly.

vslavik:
> A better fix is something I in fact started doing instead of 9487af0 before I realized the file extension is still important (this also prompted my question about getting this from API as you may recall): pass wasExportedAsXliff as another boolean argument to DownloadFile(). The logic here is complicated, so the extra verboseness and clarity would only help, and it would unambiguously say what needs to be done.

Done in 09b50afd — any thoughts?

tetatetit:
> > A better fix is something I in fact started doing instead of [9487af0](https://github.com/vslavik/poedit/commit/9487af01bb1347d25c9f756f99fc0e4b7ba2aca8) before I realized the file extension is still important (this also prompted my question about getting this from API as you may recall): pass wasExportedAsXliff as another boolean argument to DownloadFile(). The logic here is complicated, so the extra verboseness and clarity would only help, and it would unambiguously say what needs to be done.
> 
> Done in [09b50af](https://github.com/vslavik/poedit/commit/09b50afdb5ac35bf6536046c0b32dda95a6fc45b) — any thoughts?

Seem everything as discussed and expected now. Thanks a lot!

PR #648, opened 2020-07-17
`exportAsXliff: false` if original/source is XLIFF

tetatetit:
Just trying to find a reliable but quick fix for https://github.com/vslavik/poedit/pull/647#issuecomment-659878962 with Crowdin team

### DIFF
--- a/src/crowdin_client.cpp
+++ b/src/crowdin_client.cpp
@@ -425,7 +425,7 @@ dispatch::future<void> CrowdinClient::DownloadFile(int project_id,
     wxString ext(file_extension);
     ext.MakeLower();
 
-    const bool isXLIFFConverted = (ext != "po" && ext != "pot");
+    const bool isXLIFFConverted = (ext != "po" && ext != "pot" && ext != "xliff" && ext != "xlf");
 
     return m_api->post(
         "projects/" + std::to_string(project_id) + "/translations/builds/files/" + std::to_string(file_id),
--- a/src/crowdin_client.cpp
+++ b/src/crowdin_client.cpp
@@ -431,6 +431,7 @@ dispatch::future<void> CrowdinClient::DownloadFile(int project_id,
         "projects/" + std::to_string(project_id) + "/translations/builds/files/" + std::to_string(file_id),
         json_data({
             { "targetLanguageId", lang.LanguageTag() },
+            { "skipUntranslatedStrings", false },
             // for XLIFF and PO files should be exported "as is" so set to `false`
             { "exportAsXliff", isXLIFFConverted }
         }))
--- a/src/crowdin_gui.cpp
+++ b/src/crowdin_gui.cpp
@@ -825,7 +825,14 @@ void CrowdinSyncFile(wxWindow *parent, std::shared_ptr<Catalog> catalog,
                 });
 
                 if (!xliffRemoteFilename.empty())
-                    filename.Assign(xliffRemoteFilename, wxPATH_UNIX);
+                {
+                    auto ext = filename.GetExt();
+                    ext.MakeLower();
+                    if(ext == "xliff" || ext == "xlf")
+                        filename.SetExt(""); // none or any but just not XLIFF
+                    else
+                        filename.Assign(xliffRemoteFilename, wxPATH_UNIX);
+                }
 
                 return CrowdinClient::Get().DownloadFile(
                         projectId,


vslavik:
Please refrain from meaning-changing edits. They make following the discussion - or even replying! - impossible.

This is the gist of the original description here, copied over from #647:

---

> > As it comes from your #644 (comment) starting from your "For example, this Xcode-made XLIFF file differs in significant ways" and what I just tested on Crowdin to confirm - exportAsXliff: false should be used to download it "as is" just like in case with PO/POT.
> 
> I think we're moving in circles.
> 
> Again:
> 
> 1. Yes, I agree! It should behave that way. I want that.
> 2. Crowdin API has a bug currently. It doesn't work that way. Try it yourself: revert [4b68272](https://github.com/vslavik/poedit/commit/4b682728c749a28b9291cce347c9dd8769caf965), then open _Poedit Testing Sandbox →  AppSandbox.xliff_ file from Crowdin. Poedit will tell you the file is empty. It shouldn't be.
> 3. Hence applying [4b68272](https://github.com/vslavik/poedit/commit/4b682728c749a28b9291cce347c9dd8769caf965) **as a workaround for the bug** until Crowdin deploys the fix.

No! It works perfectly now with my fix #648 (https://github.com/vslavik/poedit/pull/648/commits/9d68fd54d51b6017abd06bb170779d244ca2219f). See screenshots below or just try it yourself. So it looks like Crowdin fixed that problem you mentioned related to it. Of course if I did not miss something in your explanation / description / steps to reproduce or otherwise sorry if I may be just did not delve into described your problem deeper to reveal and confirm it now fixed!

image
image

vslavik:
The original commit in this PR changed nothing; you realize it now judging from the revisions. Your screenshot is too cropped to prove anything. Perhaps you're working with a Crowdin testing environment and don't see the public API's responses?

This is how the file looks with hardcoded `bool isXLIFFConverted = false` as of now, i.e. Friday 17, 2020, 10:26 CEST:

Screen Shot 2020-07-17 at 10 11 07

This is how it looks with `bool isXLIFFConverted = true`:

Screen Shot 2020-07-17 at 10 11 44

And this is how it *should* look with `isXLIFFConverted = false` if it weren't for the bug, i.e. (postprocessed) download of the original:

Screen Shot 2020-07-17 at 10 13 55

I have highlighted the tell-tale that you can use to ID where the XLIFF came from in the latter two screenshots: one has Crowdin-assigned ID visible, one has Xcode-assigned meaningful ID.

tetatetit:
Sorry for changing description and force-pushing branch on the fly just because I realized I missed to add first and main part of proposed fix https://github.com/vslavik/poedit/pull/648/commits/4883360a2c266ce84d232fecfa73104ef7606b92#diff-ee71506cd9832717af50c2721cc35d85L428 due of some rush, sorry. So unfortunately the bug you mentioned persists and reproducable with `exportAsXliff: false` with particularly your [AppSandbox.xliff in particularly your project](https://crowdin.com/translate/poedittest/60/en-cs)
**but**
as I investigated it more and just:
1. Downloaded above file via simple `Download` (**not** `Download in XLIFF`) 
2. Added to my own test [project in Crowdin as AppSandbox (2).xliff](https://crowdin.com/translate/blablakuku/1100/en-ruby) (just shared that project with you)
3. Tried to open above file on e.g. Russian. Belarus in Poedit using build made from this PR branch just exactly **as it is right now** (i.e. with https://github.com/vslavik/poedit/pull/648/commits/4883360a2c266ce84d232fecfa73104ef7606b92#diff-ee71506cd9832717af50c2721cc35d85L428) and **it works perfectly** i.e. translations **are not empty**
6. I tried with debugger to make sure `isXLIFFConverted` is `false` so it downloaded `exportAsXliff: false`
5. Tried  "Open in Crowdin" for **your file from your project** and it still shown empty. So the only idea I have - could you please attach to this discussion your original `AppSandbox.xliff` just as you uploaded it to project tree, since I am able to get it only via `Download` in particular language

vslavik:
I fixed your access to be management and allow downloading the file from admin, but here it is too. So the implication is that this is something weird possibly file-specific?

[AppSandbox.xliff.zip](https://github.com/vslavik/poedit/files/4936914/AppSandbox.xliff.zip)

tetatetit:
Update from Crowdin team. Option `skip untranslated strings` is set in your project and that is why list of strings is empty in `exportAsXliff: false` while normal in `exportAsXliff: true` mode

vslavik:
> Update from Crowdin team. Option skip untranslated strings is set in your project and that is why list of strings is empty in exportAsXliff: false while normal in exportAsXliff: true mode

Meaning they identified the difference between our respective projects w.r.t. cause of the bug or that they consider this correct (contrary to previous communications)?

Because the current observed behavior of the "skip untranslated strings" setting doesn't make any sense to me on multiple levels:

1. doesn't affect as-is PO downloads via API — they are _always_ skipped
2. affects XLIFF as-is downloads via API, apparently
3. doesn't affect converted XLIFF — they are _never_ skipped

It does affects builds (I think), hence why it is off in my projects - otherwise all the PO files would contain non-translated filled-in junk that just takes space. But there the setting removes *translations*. It doesn't remove source text or keys. Disappearance of keys would actually be fine with gettext, but not with XLIFF where software may reasonably be doing lookup by ID w/o untranslated fallback to source language...

vslavik:
Anyway: this trivial patch seems to fix it by forcing the behavior from client side, so 🎉 hurray 🎊 I guess?

```diff
diff --git a/src/crowdin_client.cpp b/src/crowdin_client.cpp
index ef4a8db613..233c1d8262 100644
--- a/src/crowdin_client.cpp
+++ b/src/crowdin_client.cpp
@@ -431,6 +431,7 @@ dispatch::future<void> CrowdinClient::DownloadFile(int project_id,
         "projects/" + std::to_string(project_id) + "/translations/builds/files/" + std::to_string(file_id),
         json_data({
             { "targetLanguageId", lang.LanguageTag() },
+            { "skipUntranslatedStrings", false },
             // for XLIFF and PO files should be exported "as is" so set to `false`
             { "exportAsXliff", isXLIFFConverted }
         }))
```

This confuses me because the API docs say (emphasis mine)
> skipUntranslatedStrings | boolean
> Default: **false**
> Defines whether to export only translated strings

But I suppose I was misreading it, and what it *really* means is what the default project is. In any case at least as of now forcibly setting this does seem to address the issue — and doesn't affect PO downloads, where it does *not* add untranslated strings, although I think I'll push a more defensive version that does it for XLIFF only.

tetatetit:
> Anyway: this trivial patch seems to fix it by forcing the behavior from client side, so 🎉 hurray 🎊 I guess?
> 
> ```diff
> diff --git a/src/crowdin_client.cpp b/src/crowdin_client.cpp
> index ef4a8db613..233c1d8262 100644
> --- a/src/crowdin_client.cpp
> +++ b/src/crowdin_client.cpp
> @@ -431,6 +431,7 @@ dispatch::future<void> CrowdinClient::DownloadFile(int project_id,
>          "projects/" + std::to_string(project_id) + "/translations/builds/files/" + std::to_string(file_id),
>          json_data({
>              { "targetLanguageId", lang.LanguageTag() },
> +            { "skipUntranslatedStrings", false },
>              // for XLIFF and PO files should be exported "as is" so set to `false`
>              { "exportAsXliff", isXLIFFConverted }
>          }))
> ```
> 
> This confuses me because the API docs say (emphasis mine)
> 
> > skipUntranslatedStrings | boolean
> > Default: **false**
> > Defines whether to export only translated strings
> 
> But I suppose I was misreading it, and what it _really_ means is what the default project is. In any case at least as of now forcibly setting this does seem to address the issue — and doesn't affect PO downloads, where it does _not_ add untranslated strings, although I think I'll push a more defensive version that does it for XLIFF only.

Yep it really works with https://github.com/vslavik/poedit/pull/648/commits/4c784250a156b8f52cb6f9d3b759305623263510 right on the branch of this PR on [your project and file (AppSandbox.xliff) as well](https://crowdin.com/translate/poedittest/60/en-cs) so that's really 🎉 hurray 🎊, just double checking with Crowdin team if it does not have any side effect(s) 😎

tetatetit:
>  just double checking with Crowdin team if it does not have any side effect(s) 😎

Got confirmation they do not see any side effects here

tetatetit:
> 1. doesn't affect as-is PO downloads via API — they are _always_ skipped
> 2. affects XLIFF as-is downloads via API, apparently
> 3. doesn't affect converted XLIFF — they are _never_ skipped

* Crowdin team confirmed that 
  * `skipUntranslatedStrings` always overrides `skip untranslated strings`.
  *  There are still some known problem that above options (set in either of above ways) is ignored for some formats and use cases / workflows - e.g. like you mentioned above for native PO regardless of `exportAsXliff` and for **all** native formats `exportAsXliff: true`
      * Might be fixed in the future

Anyway they do not see how it could block v2.4. release with proposed by you fix/workaround https://github.com/vslavik/poedit/pull/648#issuecomment-660037991

vslavik:
I think I cleared everything remaining in d147431b + 4a1bb559 + 09b50afd, right?

tetatetit:
> I think I cleared everything remaining in [d147431](https://github.com/vslavik/poedit/commit/d147431bf69035ca9f216c89e66ff922508ef114) + [4a1bb55](https://github.com/vslavik/poedit/commit/4a1bb559408a1635af4b6a503f66eba0195eda8d) + [09b50af](https://github.com/vslavik/poedit/commit/09b50afdb5ac35bf6536046c0b32dda95a6fc45b), right?

Yep, just reviewed and everything looks matching to our discussions. Thanks a lot!

PR #653, opened 2020-07-27
Compilation fix for !HAVE_HTTP_CLIENT in 2.4

omar-polo:
Hello,

This pr wraps two call to `CanSyncWithCrowdin` inside an `#ifdef HAVE_HTTP_CLIENT`.  That functions is not available if the http client is disabled, so it wouldn't compile.  Noticed while I was updating the OpenBSD port, since there the http client is disabled (C++ rest is not yet available as package).

Lightly tested, as I'm not sure how to exactly trigger that codepath, but it shouldn't introduce regressions.

Thanks for developing this, keep up the good work :)

### DIFF
--- a/src/edframe.cpp
+++ b/src/edframe.cpp
@@ -1775,9 +1775,14 @@ void PoeditFrame::OnUpdateFromSources(wxCommandEvent&)
 
 void PoeditFrame::OnUpdateFromSourcesUpdate(wxUpdateUIEvent& event)
 {
+#if HAVE_HTTP_CLIENT
     event.Enable(m_catalog &&
                  m_catalog->HasSourcesConfigured() &&
                  !CanSyncWithCrowdin(m_catalog));
+#else
+    event.Enable(m_catalog &&
+                 m_catalog->HasSourcesConfigured());
+#endif
 }
 
 void PoeditFrame::OnUpdateFromPOT(wxCommandEvent&)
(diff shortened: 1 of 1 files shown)


AngryPenguinPL:
Hi. I ran into the similar problem when updating the package for OpenMandriva Cooker (clang compiler).

```
edframe.cpp:1780:19: error: use of undeclared identifier 'CanSyncWithCrowdin'
                   !CanSyncWithCrowdin(m_catalog));
                   ^
edframe.cpp:2542:80: error: use of undeclared identifier 'CanSyncWithCrowdin'
              if (lang.IsValid() && plForms != lang.DefaultPluralFormsExpr() && !CanSyncWithCrowdin(m_catalog))
```

Your pull request solve my problem. I don't see any problems with poedit now. Package works fine.

vslavik:
Thanks for spotting this and submitting the PR! I opted for a different fix because it is less verbose and more robust against future accidental breakages, see 29ff4b4e, but it is functionally equivalent and either is suitable for packaging patches.

omar-polo:
I think I'll take the patch from 29ff4b4 until the next release than, thanks :)

Btw, sorry for the terrible pr title!

omar-polo:
I'm sorry to bother you again, but the latest release with 29ff4b4 added on top still does not build:

```
In file included from welcomescreen.cpp:30:
./crowdin_gui.h:127:32: error: unknown type name 'CatalogPtr'
inline bool CanSyncWithCrowdin(CatalogPtr) { return false; }
                               ^
./crowdin_gui.h:128:46: error: unknown type name 'CatalogPtr'
inline bool ShouldSyncToCrowdinAutomatically(CatalogPtr) { return false; }
                                             ^
```

This is because `catalog.h` is included only if `HAVE_HTTP_CLIENT`.
```diff
--- src/crowdin_gui.h.orig
+++ src/crowdin_gui.h
@@ -26,9 +26,10 @@
 #ifndef Poedit_crowdin_gui_h
 #define Poedit_crowdin_gui_h
 
+#include "catalog.h"
+
 #ifdef HAVE_HTTP_CLIENT
 
-#include "catalog.h"
 #include "cloud_sync.h"
 #include "customcontrols.h"
 
```

vslavik:
Thank you (and sorry for the screw-up)!

Committed your patch in b551f7d572ccaed8a870ba8ad7bc779275770b43 with attribution.

omar-polo:
No problem :)

Cheers!

PR #656, opened 2020-07-30
Compilaton fix on !HAVE_HTTP_CLIENT part 2

rezso:
Don't include crowdin_gui.h from welcomescreen.cpp on !HAVE_HTTP_CLIENT

### DIFF
--- a/src/welcomescreen.cpp
+++ b/src/welcomescreen.cpp
@@ -27,12 +27,15 @@
 
 #include "colorscheme.h"
 #include "customcontrols.h"
-#include "crowdin_gui.h"
 #include "edapp.h"
 #include "edframe.h"
 #include "hidpi.h"
 #include "str_helpers.h"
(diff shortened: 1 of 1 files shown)


vslavik:
Sorry, why?

To explain the motivation for the way of fixing I choose: to minimize changes to other code. This -- adding more `#ifdefs` to other code — goes against that. 

@omar-polo's fix in https://github.com/vslavik/poedit/pull/653#issuecomment-666220171 is the right way to do it. Unless you're addressing something *else*, which is unclear from this PR or the commit message - it says what it does (which is obvious from reading the diff), but doesn't say *why*.

vslavik:
> You use an ifdef for LearnAboutCrowdinLink

I can't very well do anything else to modify the UI, can I.

> So I think, this is the simpler way.

I disagree and am not going to merge this PR. Thanks, though!

PR #666, opened 2020-11-14
Add missing include

rezso:
Fix ' concurrency.h:606:10: error: ‘atomic_bool’ in namespace ‘std’ does not name a type; did you mean ‘atomic_load’? ' build error.


vslavik:
Thank you!

PR #671, opened 2020-12-20
Fixed linker error on OSX 11.1 (Big Sur)

Janosch:
Building the master branch on Mac OSX 11.1 (Big Sur) currently fails due to a wrong library path in the gettext binaries:

Bildschirmfoto 2020-12-20 um 12 54 13

I added another call to fixup-dylib-deps.sh in generate_build_ninja.py to change the path from **/lib** to **@rpath** which fixed the problem on my machine.

I also added the Macports binary path to deps/build_deps.sh to support people using only macports (not homebrew).

Please note that Poedit still fails whith a segfault when I run it after the build, but this is a different issue.

### DIFF
--- a/deps/build.ninja
+++ b/deps/build.ninja
@@ -337,6 +337,24 @@ rule gettext_build
       rm -f $$tmp $
       && echo $
       'note: ⋯  $
+      $top_srcdir/../macos/fixup-dylib-deps.sh $
+      /lib @rpath $
+      $destdir/lib $
+      $destdir/bin/*' $
+      && $
+      tmp=`mktemp` $
+      && $
+      $top_srcdir/../macos/fixup-dylib-deps.sh $
+      /lib @rpath $
+      $destdir/lib $
+      $destdir/bin/* $
+      >$$tmp 2>&1 $
+      || (cat $
+      $$tmp ; $
+      exit 1) && $
+      rm -f $$tmp $
+      && echo $
+      'note: ⋯  $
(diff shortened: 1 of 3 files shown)


vslavik:
I'm a bit confused: the subject of this PR says it is about fixing linker errors, the screenshot shows runtime error in gettext tools, and the included commit's message talks about MacPorts only, so which is it?

Janosch:
Sorry, I am not an expert. I thought when dlyd returns an error it is called a linker error, but if you say so it is most likely a runtime error.

The error pops up independent of macports and/or homebrew. I even installed homebrew to verify that. Im pretty sure that no package manager installs libraries in /lib.

I just added the macports binaries to the PATH to enable macports users to build poedit out-of-the-box. Without this, the build fails for macports-users because ninja ist not found.

vslavik:
You're missing the point. This was (after now having reviewed the commit) a matter of [good git hygiene](https://chris.beams.io/posts/git-commit/), not the actual change or being an "expert". The commit *actually* does something completely different from what it *says* it does (and does multiple things - the above link explains well why it matters). It's these contradictory claims that are confusing.

An equivalent gettext build fix is already present in 888920d1.

I'm open to adding MacPorts support, if the PR does that one thing; ideally it would be more thorough and amend build instructions as well (see the readme and `macos/Brewfile` files for what is needed).

Janosch:
Ok. Got it.

PR #685, opened 2021-02-22
Escape '&' in the comment sidebar

lanurmi:
Fixes the following issue:
If you have a comment containing an ampersand (&) in the comment
sidebar, it would be treated as an accelerator, i.e. the ampersand
would not be shown, and the character after it would be underlined
(on Windows), or not (on Mac).


vslavik:
Thank you!

vslavik:
I think this is actually a much wider issue than just the comments -- *any* wrapped text in the sidebar or elsewhere, including the suggestions, is affected.

My first reaction was that using `SetLabelText()` would do the trick too, but then realized your fix is better and more robust, because it works for all uses and doesn't require us to hunt them all down.

Committed now. Thanks again!

Issue #230, opened 2015-12-16
Colorization of current/new msgid for fuzzy translations


sebras:
These are two wishlist items.

I'm currently using poedit 1.8.6 for doing translations into Swedish. For fuzzy translations you can currently see the dequoted string of the current msgid in the source text field while the the old msgid is shown in the side bar as a quoted string.

First of all I'd prefer the old msgid also to be a dequoted string (without the preceeding msgid and " " for each line).

Secondly, attempting to figure out what part of a long msgid has changed is currently quite difficult. Having some form of color coding for the parts that changed in a fuzzy string would be useful. I'm thinking of something along the lines of red for removed parts (in the sidebar) and green for added parts (in the source text field). Similar to the deep green and deep red in the micro version number in [this commit](https://github.com/vslavik/poedit/commit/26f56e664e3911460d70766b3243b65bd2410bc3?diff=split) here on github. It is quite likely I'm asking you to implement diff(1) with coloring which might be a bit difficult. I guess if the differences are too big then poedit could just claim that the entire old msgid was removed and that the entire current msgid was the replacement. It seems like the current implementation uses wxControl::SetText() while probably wxControl::SetLabelMarkup() could be used to achieve the coloring. Do you think this idea is feasible, and it is something you would consider?

poedit is an excellent tool and it has certainly helped me a lot when doing translations!
Thanks for the effort you put into the project! Even if you may decide to skip this wishlist item. :)

vslavik:
> These are two wishlist items.

[ Please don't mix unrelated things in a single issue — you’re making it impossible to handle it sanely (when do you close such an issue? with what commit?) ]

> Secondly, attempting to figure out what part of a long msgid has changed is currently quite difficult. 

Your problem is not the lack of msgid diffing in Poedit. Your problem is *that you have too long msgids*, which isn’t normal for software localization — I suspect gettext misuse for something like freeform text, am I right? 

You’re not the only one asking about something like this, so I’m leaving this open. But honestly, it’s a rather obscure requirement that doesn’t normally arise in localization. I’m not likely to address it in immediate (or even not so immediate) future. A PR (based e.g. on https://code.google.com/p/google-diff-match-patch/) would go a long way.

sebras:
[That's a fair comment, I'll separate out the dequoting into a separate issue. Sorry for the inconvenience. :)]

> Your problem is that you have too long msgids, which isn’t normal for software localization — I suspect
> gettext misuse for something like freeform text, am I right?

[...]

PR #687, opened 2021-02-25
preliminary Diff implementation for #230 (draft)

Fat-Zer:
TODO:

* [ ] `wxStaticText` resizing (and text rewrapping) should be reimplemented.
* [ ] Add settings for colour of diff text background in UI (people with dark UI theme will likely want to change those).
* [ ] Add context menu option to copy old source text.

Note: It assumes dtl is installed system-wide during build. Feel free to change it if you are going to package it with tarballs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
     - uses: actions/checkout@v4
     - name: Checkout submodules
       # submodules are not used by the build, but needed for `make dist`:
-      run: git submodule update --init deps/json deps/pugixml
+      run: git submodule update --init deps/json deps/pugixml deps/dtl
     - name: Install dependencies
       run: |
         sudo apt-get update && \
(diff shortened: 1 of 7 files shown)


vslavik:
OK, I think I give up on trying to make static text work; as referred to in the other issue, the only path forward seems to be to use read-only wxTextCtrl as you suggested initially.

cooljeanius:
does this support `msgmerge`'s `--previous` flag?

cooljeanius:
My previous comment was not off-topic; `msgmerge`'s `--previous` flag is an existing way for marking differences between translation strings that should be considered when implementing a new diff implementation, such as this PR attempts to do.

Fat-Zer:
rebase upon current master and bundle dtl into tarball (there is little reason in keeping source-only library separated)

Neustradamus:
@Fat-Zer, @vslavik: Any progress about this PR?

PR #716, opened 2021-09-30
Normalize order of PO header data lines

lanurmi:
Ensure that the lines of a PO header are written in the same canonical
order as in a header generated by xgettext or rearranged by msgmerge.

The actual issue that this commit fixes is: When you open a .po whose
header doesn't contain a "Language:" line, such a line will be written
at save, but it will end up after the "Plural-Forms:" line. The proper
place for it is after "Language-Team:".

Such variation is not dangerous, but causes unnecessary changed lines
in commits, if a file is sometimes last modified by Poedit, sometimes by
msgmerge.


vslavik:
Perfect, thanks a lot for doing this!

Issue #703, opened 2021-06-07
icon-theme.cache being packaged


rffontenelle:
`/usr/share/icons/hicolor/icon-theme.cache` is being packaged together with other relevant Poedit files during `make install` step. This file is generated when running `gtk-update-icon-cache`, which distros usually run on post-install phase of package installation, and usually it is not packaged together.

If possible to not install this file, I haven't found. Please enlighten if there is a well and I'll pass it downstream

Poedit version: 3.0
Distro: Arch Linux 64-bit
Package: [poedit 1:3.0-1](https://archlinux.org/packages/community/x86_64/poedit/)

I experienced this issue with the package provided in arch official repos, and then I built it myself to achieve the same result. I see the following output when installing the package:

```
$ sudo pacman -S poedit
resolving dependencies...
looking for conflicting packages...

Packages (1) poedit-1:3.0-1

Total Installed Size:  6.77 MiB

:: Proceed with installation? [Y/n] 
(1/1) checking keys in keyring                     [######################] 100%
(1/1) checking package integrity                   [######################] 100%
(1/1) loading package files                        [######################] 100%
(1/1) checking for file conflicts                  [######################] 100%
error: failed to commit transaction (conflicting files)
poedit: /usr/share/icons/hicolor/icon-theme.cache exists in filesystem
Errors occurred, no packages were upgraded.
```

Arch's package manager prints this error when the package is trying to replace existing file that does not belongs the package itself (e.g. in previous version).

vslavik:
`make install` does not "package" anything - that's distribution's job. 

`make install` *installs* Poedit and in doing so, updates the cache as [it is supposed to](http://ptomato.github.io/advanced-gtk-techniques/html/desktop-file.html).

Please direct complains about packaging to those responsible for the packages, as that is what is broken here, not Poedit's makefiles.

c72578:
@rffontenelle Remark: In Fedora we remove the installed `icon-theme.cache` file before packaging.
e.g.: `rm %{buildroot}/usr/share/icons/hicolor/icon-theme.cache`
See: https://src.fedoraproject.org/rpms/poedit/c/dc4deb47ab540bf6524b0574ae2e9c51e17638bd?branch=rawhide

rffontenelle:
@c72578 Yeah, this solution was my initial thought but I since I've never saw a upstream software installing icon-theme.cache, I thought it would worth reporting upstream as this could be an issue for many distros. I'll report this to downstream too.

[...]

PR #732, opened 2022-01-30
Do not install icon cache upon staged install

super7ramp:
Upon installation, updating icon cache is relevant so that new icons can be found by the system.

Upon a [staged installation](https://www.gnu.org/software/automake/manual/html_node/Staged-Installs.html) though, i.e. when `DESTDIR` is set to a staging installation area, it is not relevant to update the icon cache as this cache is not the actual target system's cache.

Staged installation is typically used when creating a package of the application. Cache update is a problem well known to packagers which typically handle it by wrapping cache update commands into scripts to be executed at post (un)installation time - on the real target system.

See also:
- #703
- [The tutorial](https://ptomato.name/advanced-gtk-techniques/html/desktop-file.html) referenced in the ticket which actually mentions this tricky staged install case

### DIFF
--- a/artwork/Makefile.am
+++ b/artwork/Makefile.am
@@ -48,7 +48,7 @@ dist_uiicons_DATA = \
 		window-close.png
 
 update-icon-cache:
-	which gtk-update-icon-cache >/dev/null && gtk-update-icon-cache -f -t $(DESTDIR)$(iconsdir) || true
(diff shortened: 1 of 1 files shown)


vslavik:
Thank you for setting me straight (I misunderstood the purpose of this bit), and thank you for the fix!

PR #742, opened 2022-03-16
Change git URLs to HTTPS in submodules

tsunamistate:
Unencrypted git:// protocol is no longer supported and errors out when trying to pull `letsmove`, `LucenePlusPlus` and `zlib` submodules
More info here: https://github.blog/2021-09-01-improving-git-protocol-security-github/


vslavik:
Thanks a lot!

P.S. You'd think GH would scan affected repositories and send an email about it...

Issue #714, opened 2021-09-06
The "ID" section takes up a lot of space


eshagh79:
Hi
When the "Show ID" option is enabled, the ID field takes up a lot of space, leaving no space for the "Translate" and "Text Source" sections.

Reproduction:
1. Open the poedit application with a po file (any file)
2. From the toolbar, click the "View" button, and select the "Show String ID" option.
3. Resize poedit (resize too much)
Note: The "ID" field is not displayed before resizing (raised in another issue)

Additional information:
I am using Ubuntu 21.04.
poedit version 3.0-6
My computer language is Persian (RTL direction), but these problems exist even when the program is opened in English. 

I tried both window system (wayland) and (X11) and the problem persisted.
![نماگرفت از 2021-09-06 18-06-29](https://user-images.githubusercontent.com/52137191/132234656-0245e21a-f94a-4a68-aadb-0fd40b41dae4.png)
![نماگرفت از 2021-09-06 18-14-12](https://user-images.githubusercontent.com/52137191/132234666-6f0286bc-f82c-46dd-baf1-78590bf35bb1.png)

vslavik:
Please try the version from https://poedit.net and verify whether it's an issue with Poedit or Ubuntu packaging ("3.0-6" is evidently an Ubuntu package version, even though you don't say so).

eshagh79:
Hi again.
I installed the latest version of poedit from snap. And I can still reproduce the problem.
latest / stable: 3.0 2021-06-07 (30) 100MB -

eshagh79:
Although in is not related to this problem, but I have improved some Persian translation in crowdin if you can add it to poedit

vslavik:
> Although in is not related to this problem,

If you find yourself writing that in an issue, stop and don't hit the Comment button.

vslavik:
Please don't close issues without explanation. It's rude after you asked us to spend time on it, and is counterproductive.

super7ramp:
I can reproduce the issue with the following environment:

- Poedit version: 3.0.1
- wxGTK version: 3.1.6
- GTK+ version: 3.24.33
- Operating system: openSUSE Tumbleweed (20220414)

I think the issue comes from the fact that, at least on wxGTK, last `wxDataViewColumn`:

- Can grow despite its width being fixed
[...]

Issue #643, opened 2020-07-06
ID column doesn't appear until window is resized


Spixmaster:
**Environment**
Operating System: Arch Linux
KDE Plasma Version: 5.19.2
KDE Frameworks Version: 5.71.0
Qt Version: 5.15.0
Kernel Version: 5.7.7-arch1-1
OS Type: 64-bit
Processors: 8 × Intel® Core™ i5-8265U CPU @ 1.60GHz
Memory: 7.7 GiB of RAM
Graphics Processor: Mesa Intel® UHD Graphics 620
DE: Plasma
WM: KWin

**Bug**
The string ID on the right does not appear if a .po file is just opened. It only appears if the window is resized to a smaller resolution and then resized to full screen.

vslavik:
Quoting another report by Antonio Rosella:

I'm using poedit 2.4.2 on an MXLInux 19.2 distribution ( an AntiX-Debian derivative, XFCE desktop https://mxlinux.org/ ). The package came from the distribution stable repository

Linux htpc 5.9.0-3mx-amd64 #1 SMP Debian 5.9.9-1~mx19+1 (2020-11-21) x86_64 GNU/Linux

Here attached you can find 3 snap that hopefully describe the issue.
The first (poedit_1.jpg) is the opening window that pop up at the program startup: as you can notice, column "ID" it is not present at all, even if "View -> Show String ID" is selected

Now, if I enlarge the window on the right side, the ID colum show itself. (poedit_2.jpg) . You have to resize the full windows: if you try to resize only the "Translation
suggestions" subwindow nothing happen

Again, reducing the window size, i can be noticed that the column "ID" start to resize, until it disappear completely (image poedit_3.jpg is a snap in the middle of the process...). It is at this point that the massages I sent you pop up on the console, arguing about "Negative content width -7 (allocation 5, extents 6x6) while allocating gadget (node button, owner GtkButton)"

It seems that on shrink of the window, the column "ID" start to reduce his size further than allowed

super7ramp:
I can reproduce the issue with the following environment:

- Poedit version: 3.0.1
- wxGTK version: 3.1.6
- GTK+ version: 3.24.33
- Operating system: openSUSE Tumbleweed (20220414)

I think the issue comes from the fact the sizing of the ID column is not immediate - at least on wxGTK [^note].

So in `PoeditListCtrl::UpdateColumns`:

```c++
    if (m_displayIDs)
    {
[...]

PR #747, opened 2022-04-24
Fix sizing of the ID column on wxGTK

super7ramp:
This is a naive attempt to fix the sizing issues of the ID column observed with wxGTK and reported in #643 and #714.

Basically:

- Sizing computation is delayed, so we can't fix the width immediately until we're sure it has been computed
- Despite being fixed, last column automatically grows and `wxDataViewColumn::GetWidth()` returns the actual width, not the fixed width, so we need a way to memorize the fixed width ourselves. I've created a custom `wxDataViewColumn` for that, I'm not sure it's the best approach though

I've tested this with the following environment:

- wxGTK version: 3.1.6
- GTK+ version: 3.24.33
- Operating system: openSUSE Tumbleweed (20220414)

### DIFF
--- a/src/edlistctrl.cpp
+++ b/src/edlistctrl.cpp
@@ -177,10 +177,12 @@ public:
      */
     int GetWidth() const override
     {
+#if !wxCHECK_VERSION(3,1,3)
         // workaround a wx bug where it calculates width of hidden columns
         // see https://github.com/wxWidgets/wxWidgets/commit/560a81b913f23800e286d297d8cd38e72a207641
         if ( IsHidden() )
             return 0;
+#endif
 
         if ( fixed_width != wxCOL_WIDTH_DEFAULT )
             return fixed_width;
--- a/src/edlistctrl.cpp
+++ b/src/edlistctrl.cpp
@@ -156,6 +156,62 @@ public:
 
 #endif
 
+/**
+ * A wxDataViewColum whose GetWidth() prefers to return the fixed width over the actual width - they may differ, e.g.
+ * for the last column which may automatically grow to take the available space on resize.
+ */
+class DataViewFixedColumn : public wxDataViewColumn
+{
+
+public:
+
+    DataViewFixedColumn(const wxString &title, wxDataViewRenderer *renderer,
+                       unsigned int model_column, int width = wxDVC_DEFAULT_WIDTH,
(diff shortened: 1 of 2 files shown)


vslavik:
> I don't follow the reasoning or the code either, but it may be that I just need one more fresh look.

Never mind, the explanatory notes in #643 and #714 help - thanks!

super7ramp:
> I don't understand the reference to https://github.com/wxWidgets/wxWidgets/commit/560a81b913f23800e286d297d8cd38e72a207641 in code

This special handling of the width value when column is hidden existed in a different form: It has been moved from [line 730](https://github.com/vslavik/poedit/pull/747/commits/f2b5fef69a6e734737600bbd09bd2f4dccbc8569#diff-c2d0a6672c1362a31f790b0f82da4d5cac47e97cd964fec97c75492fd212032fL730). Instead of `PoeditListCtrl::CreateColumns()` setting the width to 0 when column is hidden, I let `DataViewFixedColumn::GetWidth()` returning 0 when it is hidden - which seems functionally equivalent.

This was not a necessary change and lacked explanations. It felt to me the new `DataViewFixedColumn::GetWidth()` was a more appropriate place to handle this specific handling of the ID column width than `PoeditListCtrl::CreateColumns()`.

Should I:

- Revert to previous way of doing? or
- Split the change into two commits to explain the move?

vslavik:
> This special handling of the width value when column is hidden existed in a different form: It has been moved from [line 730](https://github.com/vslavik/poedit/pull/747/commits/f2b5fef69a6e734737600bbd09bd2f4dccbc8569#diff-c2d0a6672c1362a31f790b0f82da4d5cac47e97cd964fec97c75492fd212032fL730).

Yes, but you're referring a specific bug in wx here, one which was already fixed. If this part of the code is for compatibility with older wx, then it should be inside `#if !wxCHECK_VERSION(3,1,6)` so that it can be removed once wx-3.2 is a prerequisite (which it will be the moment a stable 3.2 series is released; I'm only supporting wx-3.0 compilation to accommodate conservative distros).

vslavik:
Just a note that this fix doesn't address a similar problem on macOS 12 - that's to be expected, but I hoped it could. I'm going to merge this together with macOS fix - it's possible (likely?) that one will affect the other...

vslavik:
The macOS issue was unrelated (111df0f3); both are now merged.

Thanks a lot for figuring the tricky issue under wxGTK and implementing the fix!

Issue #758, opened 2022-06-13
Error: assert ""IsOk()"" failed in GetWidth(): invalid bitmap


libreajans:
Hi, I installed poedit from Flathub via flatpak

`flatpak install flathub net.poedit.Poedit`

When I run it, it has two errors
`
flatpak run net.poedit.Poedit`

```
Gtk-Message: 21:07:29.251: Failed to load module "canberra-gtk-module"
Gtk-Message: 21:07:29.252: Failed to load module "canberra-gtk-module"
Debug: Failed to connect to session manager: None of the authentication protocols specified are supported
./src/gtk/bitmap.cpp(937): assert ""IsOk()"" failed in GetWidth(): invalid bitmap
Debug: window wxTextCtrl(0x55912da24630) lost focus even though it didn't have it
Debug: window wxTextCtrl(0x55912da24630) lost focus even though it didn't have it


(net.poedit.Poedit:2): Gtk-CRITICAL **: 21:07:38.784: gtk_box_gadget_distribute: assertion 'size >= 0' failed in GtkScrollbar
(net.poedit.Poedit:2): Gtk-CRITICAL **: 21:07:38.820: gtk_box_gadget_distribute: assertion 'size >= 0' failed in GtkScrollbar

Debug: cannot read address information for stack frame #0
./src/gtk/bitmap.cpp(767): assert ""IsOk()"" failed in ConvertToImage(): invalid bitmap
./src/common/image.cpp(438): assert ""IsOk()"" failed in Scale(): invalid image
./src/gtk/bitmap.cpp(593): assert ""image.IsOk()"" failed in wxBitmap(): invalid image

(net.poedit.Poedit:2): Gtk-CRITICAL **: 21:07:50.489: gtk_box_gadget_distribute: assertion 'size >= 0' failed in GtkScrollbar
Debug: window wxTextCtrl(0x55912da24630) lost focus even though it didn't have it
Debug: window wxTextCtrl(0x55912da24630) lost focus even though it didn't have it

```
![Ekran Görüntüsü - 2022-06-13 21-10-14](https://user-images.githubusercontent.com/6218679/173417576-d0a23e5c-ea15-449e-9697-07dd32062483.png)

![Ekran Görüntüsü - 2022-06-13 21-10-22](https://user-images.githubusercontent.com/6218679/173417573-b1647626-268d-44dd-8493-824102a2983c.png)

libreajans:
Poedit version: 3.1

[...]

PR #759, opened 2022-06-15
Do not try to rescale recent file icon when it cannot be loaded

super7ramp:
Recent file icon loading may fail e.g. because of missing read
permission or other packaging issue.

Icon rescale code introduced by 5fc8d6c did not foresee this
possibility.

Add a condition to prevent it from manipulating a non-ok icon.

This prevents assertion error although icon loading failure will still
lead to an error message showed to user.

See #758.


vslavik:
Thanks for the fix! I should have caught this myself in review, but completely missed it too.

PR #771, opened 2022-10-10
Skip files that do not exist

dmitryrechkin:
Fixes the issue for cases when you are trying to update from source code and one of the files is a link to non-existing, which led to Permission denied exception.


vslavik:
Thank you! I added an explanatory comment about the additional check and merged - will be in Poedit 3.2 very soon.

PR #772, opened 2022-10-22
Fix chmod/chown return code verification logic

super7ramp:
On Linux, `TempOutputFileFor::ReplaceFile` returns a wrong status, indicating the replacement has failed when it has not. This leads to a misleading error message displayed to user on every po file save.

The cause is in the verification of the return codes of `chown`/`chmod`. `chown` and `chmod` return a non-zero integer value when an error occurs. A non-zero value is considered as a boolean `true`.

This commit fixes the logic error; It explicitly compares the returned integer value to another integer to avoid confusion.

### DIFF
--- a/src/utility.cpp
+++ b/src/utility.cpp
@@ -302,9 +302,9 @@ bool TempOutputFileFor::ReplaceFile(const wxString& temp, const wxString& dest)
   #ifdef __UNIX__
     if (overwrite)
     {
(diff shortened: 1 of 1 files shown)


vslavik:
Introduced in 91c5f13257ffed9b80dc6744546f9cd25a2725ae :(

Thanks, this is serious enough to warrant immediate 3.2.1

PR #801, opened 2023-08-13
qa_checks: handling of all upper-case

GitMensch:
done per "re-request"

Note: because I'm not familiar with the building or the test process: this is untested an likely should get a check in the internal testsuite.


vslavik (changes requested):
In addition to per-line issues, there's also

1. "this is untested" is a red flag, means this PR was premature and explains some of the confusing/surprising things in there
2. the commit message is unclear and non-explanatory; please follow standard practice: https://cbea.ms/git-commit/

vslavik:
Closing abandoned PR.

PR #804, opened 2023-08-26
Add accessibility support to SwitchButton on Windows

### LINKED ISSUE #693: The state of the "Need work" toggle button is not reported through accessibility APIs
This Checkbox is announced as a button using NVDA and Windows Narator. Buttons are pressed typically with enter which isn't working. Pressing Space to switch works like a checkbox. It would be great if you'll make sure that switching is announced by nvda or narator.

### LINKED ISSUE #696: Prepend "needs work" to the accname for fuzzy strings in the list
Since rewrite with WxWidgets we have lost easy identification of strings requiring work when browsing the list.
Back in the days of win32 list view we used to check if the text is bold at screen reader level. With Wx this is no longer available to do.
For displaying the list poedit now has custom list control with custom accessibility related tweaks.
Would it be doable to prepend accessible name of each list item that has fuzzy string with a string fuzzy, needs work or something similar please?

leonardder:
Fixes #693 

### Issue summary
There is no reliable way for screen reader users to find out the fuzzy (needs work) status. After some research, the most likely cause is as follows:

1. SwitchButton is owner drawn and therefore sets the BS_OWNERDRAW window style
2. When BS_OWNERDRAW is set, all other window styles should be ignored [according to the docs](BS_OWNERDRAW)
3. The default MSAA implementation in Windows ignores other window styles. As it relies on window style to find out whether the button is either a push button or a check box, this information is lost, and MSAA treats the toggle as a plain button without checkable state.

### Solution
Implement a wxAccessible subclass as outlined in #696 

### Testing performed
Executed from source, ensured that the needs work button is reported as check box and that toggling the button is properly reported by the NVDA screen reader.

### DIFF
--- a/src/custom_buttons.cpp
+++ b/src/custom_buttons.cpp
@@ -413,7 +413,7 @@ SwitchButton::SwitchButton(wxWindow *parent, wxWindowID winid, const wxString& l
     Bind(wxEVT_LEFT_DOWN, &SwitchButton::OnMouseClick, this);
 #if wxUSE_ACCESSIBILITY
     Bind(wxEVT_TOGGLEBUTTON, [=](wxCommandEvent& e) {
-        wxAccessible::NotifyEvent(wxACC_EVENT_OBJECT_STATECHANGE, wxDynamicCast(this, wxWindow), wxOBJID_CLIENT, wxACC_SELF);
+        wxAccessible::NotifyEvent(wxACC_EVENT_OBJECT_STATECHANGE, this, wxOBJID_CLIENT, wxACC_SELF);
         e.Skip();
     });
 #endif // wxUSE_ACCESSIBILITY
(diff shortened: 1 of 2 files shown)


leonardder:
@vslavik Regarding code style, is there a style guide somewhere? I started with visual studio, then realized overall code style is very different from the current style of poedit.

vslavik:
Thanks! 

> @vslavik Regarding code style, is there a style guide somewhere?

Just mimic the surrounding code... I'm guessing you ask because of class naming and the somewhat schizophrenic CamelCase vs cxx_standard_like? I started switching to the latter in mostly non-GUI or abstracted code, but not sure it was such a bright idea, as it seems weird around wxWidgets' GUI code...

This looks good to me, I just pushed a removal of unneeded `dynamic_cast` in 4d334d127e71f4847a9fa7213395fb604832c7ee.

vslavik:
> There is no reliable way for screen reader users to find out the fuzzy (needs work) status. After some research, the most likely cause is as follows:

I forgot to address this: that does sound plausible, thanks for making sense of the situation.

Merged now, thanks again!

PR #841, opened 2024-05-04
Add astro to supported extensions for Javascript

Suven:
Hey!

Thanks for Poedit (: I am using the proversion and am currently in the process of writing a skeleton / tutorial for how to setup a SEO friendly multilingual project with Astro and po-files.

Astro is a very quickly growing and already big framework for creating SSG'd websites. It's pages have an extension of `.astro` but contain mainly js / tsx.

My tutorial will feature how to do the basic setup with gettext and show how Poedit can be used for the actual translation. It's nearly done and I will link it within this PR soonish.

It would be cool if I could explain how to use poedit, without needing to require profeatures to beginners. I think the only missing part for that is the extension-support.

Edit: Here is the project I was talking about https://suven.github.io/astro-po-i18n/

### DIFF
--- a/scripts/extract-fileviewer-mappings.py
+++ b/scripts/extract-fileviewer-mappings.py
@@ -22,6 +22,7 @@ OVERRIDES = {
     'h'     : 'cpp',
     'inc'   : 'php',
     'cake'  : 'coffeescript',
+    'astro' : 'javascript',
     'es'    : 'javascript',
     'fcgi'  : 'lua',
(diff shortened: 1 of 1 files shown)


vslavik:
This PR only changes the fileviewer, not anything about actual extraction, i.e. is clearly not tested.

This does not seem to be standard localization for Astro, so I don't think this is appropriate to merge, sorry. You have the option to customize file extensions in Poedit yourself for edge cases like this.

PR #850, opened 2024-07-19
Use static IDs for controls needed by NVDA

vslavik:
This reorganizes Poedit's code to be able to a) assign static IDs safely and b) add more quickly if the need arrises.

See the relevant NVDA plugin's code here: https://github.com/nvaccess/nvda/blob/master/source/appModules/poedit.py

@LeonarddeR I'd appreciate your review if you could spare the time. Most of the changes are internal to Poedit and of little interest, the interesting bit is the `static_id.h` file:

1. Is it understandable?
2. Does it cover NVDA's needs?
3. Can it be improved and/or is there more that you guys need?

I *think* it covers it (`PRO_IDENTIFIER` and `MAIN_SPLITTER_IDENTIFIER` seem to be needed just for the clever/hacky workaround for lack of static IDs), but I'm far from being 100% sure.

### DIFF
--- a/Poedit.vcxproj
+++ b/Poedit.vcxproj
@@ -241,6 +241,7 @@
     <ClInclude Include="src\recent_files.h" />
     <ClInclude Include="src\sidebar.h" />
     <ClInclude Include="src\spellchecking.h" />
+    <ClInclude Include="src\static_ids.h" />
     <ClInclude Include="src\str_helpers.h" />
     <ClInclude Include="src\syntaxhighlighter.h" />
     <ClInclude Include="src\text_control.h" />
--- a/Poedit.vcxproj.filters
+++ b/Poedit.vcxproj.filters
@@ -425,6 +425,9 @@
     <ClInclude Include="src\app_updates.h">
       <Filter>Header Files</Filter>
     </ClInclude>
+    <ClInclude Include="src\static_ids.h">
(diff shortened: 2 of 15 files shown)


LeonarddeR:
@vslavik I'm really grateful for your kind work, I'm pretty sure this will cover everything we need.
Would it be possible to provide a build of this code others can test too? Preferably I'd like to prototype a poedit.py with the hardcoded control ids to test with so we're absolutely sure that this settles everything we need.

vslavik:
> Would it be possible to provide a build of this code others can test too?

@LeonarddeR Of course!

https://download.poedit.net/tmp/Poedit-3.5nvda-setup.exe

LeonarddeR:
I filed https://github.com/nvaccess/nvda/pull/16889 and tested the build, it seems to work great up until now.

vslavik:
> I filed https://github.com/nvaccess/nvda/pull/16889 and tested the build, it seems to work great up until now.

OK, thanks for testing, I plan to merge it in time for 3.5 (Real Soon Now™).

LeonarddeR:
Thanks @vslavik.
Is there a due date for the release of poedit 3.5?

vslavik:
@LeonarddeR Well, I wanted it to be out for a while now, so... 🤷‍♂️ This month for sure.

PR #901, opened 2025-06-05
Fix not being able to view source files without a line number

### LINKED ISSUE #900: Can't view source file if no line number is specified
## Info
Version: Poedit v 3.6.3 from poedit.net
OS: Windows 10

## Description
When a string in a po file has a linked source code path without a line number `#: src/main.gd`, trying to view the source code via the "Code occurrences" option in the right click menu, Poedit will say 
```
Source code not found
src/main.gd
Poedit cannot show source code where the string is used, because the file is either not available in the referenced location or it is a symbolic reference that doesn’t point to a real file..

```
Thats because the linked file path doesn't include a colon and the returned string from[ `BeforeLast()` ](https://docs.wxwidgets.org/stable/classwx_string.html#ae7538d32a43f35e065736750d4b169de) will be empty.

https://github.com/vslavik/poedit/blob/c4fe890ef72c8c8cbc6b9b8cc1784dba10447798/src/fileviewer.cpp#L201

When working around this by adding colons to the ends of all file paths, poedit opens them without issues even with missing line numbers.

## Video

https://github.com/user-attachments/assets/07e9cfc4-7fed-4a23-9d0e-8fe4ec8c7942

## Solution
Poedit should be able to view source code files even when the path doesn't specify a line number.

Kiisu-Master:
Closes #900

### DIFF
--- a/src/fileviewer.cpp
+++ b/src/fileviewer.cpp
@@ -199,13 +199,9 @@ wxFileName FileViewer::GetFilename(wxString ref) const
 
     wxPathFormat pathfmt = ref.Contains(_T('\\')) ? wxPATH_WIN : wxPATH_UNIX;
 
-    wxString filepath = ref;
-    wxString without_linenumber = filepath.BeforeLast(_T(':'));
-    if ( !without_linenumber.IsEmpty() )
-    {
-        filepath = without_linenumber;
-    }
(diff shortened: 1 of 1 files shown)


vslavik:
Thanks!

PR #910, opened 2025-09-08
Suppress false warning for uppercase translation when source is "translator-credits"

KristjanESPERANTO:
"translator-credits" (commonly used, for example, in GNOME projects) is typically filled with proper names (starting uppercase).

### DIFF
--- a/src/qa_checks.cpp
+++ b/src/qa_checks.cpp
@@ -218,12 +218,13 @@ public:
 
         if (u_islower(source[0]) && u_isupper(translation[0]))
         {
-            if (m_lang != "de")
+            if (m_lang != "de" && source != "translator-credits")
             {
                 item->SetIssue(CatalogItem::Issue::Warning, _("The translation should start with a lowercase character."));
                 return true;
             }
-            // else: German nouns start uppercased, this would cause too many false positives
+            // Suppress warning for German, because nouns start uppercased and this would cause too many false positives.
+            // Also suppress for 'translator-credits', which is conventionally filled with proper names (starting uppercase), not translated text.
         }
 
         return false;


KristjanESPERANTO:
> This feels too special-cased... Can you please provide some evidence for widespread common use?

I encounter this frequently while translating GNOME projects. Many of them include "translator-credits" (e.g. Nautilus: https://l10n.gnome.org/POT/nautilus.main/nautilus.main.es.po?view=co or GIMP: https://l10n.gnome.org/POT/gimp.master/gimp.master.es.po?view=co).

Additionally a GitHub code search shows 166k+ occurrences (https://github.com/search?q=%22msgid+\%22translator-credits\%22%22&extension=po&type=code). So I think it's not too "special-cased" :slightly_smiling_face: 

> Also doesn't make sense to do for German only.

Yes, hard-coded exceptions aren’t pretty; alternatively, this could become a user option. But I’m not very experienced in C++, though.

vslavik:
> > Also doesn't make sense to do for German only.
>
> Yes, hard-coded exceptions aren’t pretty; alternatively

I didn’t say that “hardcoded exceptions aren’t pretty”.  I said that *your PR does this check for German only, which doesn’t make any sense*.

(OK, I didn’t say explicitly the part about what *your* code does, because I assumed that *you* know what *you* wrote.)

In its current form, this PR is unmergeable.

KristjanESPERANTO:
> your PR does this check for German only,

The exception for German was already in the code. I just added another exception.

> In its current form, this PR is unmergeable.

Okay, what do you suggest?

KristjanESPERANTO:
> So, only GNOME and not not even all of it?

Yes. It does not appear to be an official standard, and it is not easy to find proper documentation on it. So, maybe you're right that it is too special-cased.

> You're disabling one check only, and mangle it together with a check for German in a way that is confusing and hard to maintain.

Okay, I understand that. I looked for a simple way to do it.

Let's drop the PR, I've been able to ignore this wrong warning in recent years, so I can continue to do so.

----

Thanks for poedit by the way, I'm using it for ages! :slightly_smiling_face:

vslavik:
> I looked for a simple way to do it.

"Simple" is good. "Just throw it at some random place" is not.

> Let's drop the PR,

If you're not prepared to work on fixing your PRs' issues, don't create them - you're asking maintainers to spend time on something you want, and it is rude to waste their time with drive-by PRs you yourself don't care for :/

